PR #43, opened 2019-01-23
Add -nometadata to go-bindata to stop updating generated files by every build

nak3:
This patch adds `-nometadata` to go-bindata option to stop updating metadata by every build.


nak3:
@wongma7 @jsafrane Could you please take a loot? This is same way with [cluster-storage-operator](https://github.com/openshift/cluster-storage-operator/blob/ed009bb62d3834d3968aedbed91621c85b9e5d8f/Makefile#L21)

PR #45, opened 2019-01-23
Fix golint error

nak3:
Currently golint (`./hack/verify-golint.sh `) fails due to following error:

```
$ ./hack/verify-golint.sh 
!!! golint detected the following problems:
./pkg/controller/csidriverdeployment/objects_test.go:19:25: should omit type corev1.Container from declaration of var defaultDriverContainer; it will be inferred from the right-hand side
./pkg/controller/csidriverdeployment/objects_test.go:26:12: should omit type v1alpha1.CSIDriverDeployment from declaration of var defaultCR; it will be inferred from the right-hand side
```

This patch fixes the error.

### DIFF
--- a/pkg/controller/csidriverdeployment/objects_test.go
+++ b/pkg/controller/csidriverdeployment/objects_test.go
@@ -16,14 +16,14 @@ import (
 )
 
 var (
-	defaultDriverContainer corev1.Container = corev1.Container{
+	defaultDriverContainer = corev1.Container{
 		Name:  "defaultDriverContainer",
 		Image: "defaultDriverImage",
 	}
 
 	sixty = int32(60)
 
-	defaultCR v1alpha1.CSIDriverDeployment = v1alpha1.CSIDriverDeployment{
+	defaultCR = v1alpha1.CSIDriverDeployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "default",
 			Namespace: "default",


jsafrane:
go fmt changes between 1.10 and 1.11. Last time I checked we used 1.10, so anything you push should be formatted with 1.10.

nak3:
@jsafrane Thank you! updated (and it passed the Jenkins test.)

nak3:
Ah, I misunderstood. The format error fixed in this PR causes error with go1.11, but go 1.10 misses this error. So, verify test passes either with or without this.

nak3:
@jsafrane I'm sorry for bothering you. But I realized that this is an issue of `golint` and whenever I run `make verify` or `hack/verify-golint.sh`, the script always gets error.

```
$ ./hack/verify-golint.sh 
!!! golint detected the following problems:
./pkg/controller/csidriverdeployment/objects_test.go:19:25: should omit type corev1.Container from declaration of var defaultDriverContainer; it will be inferred from the right-hand side
./pkg/controller/csidriverdeployment/objects_test.go:26:12: should omit type v1alpha1.CSIDriverDeployment from declaration of var defaultCR; it will be inferred from the right-hand side
```

Can I ask you what golint version do you use? I tested on my older and newer Fedora (golint-0.0.x) but both fails.

jsafrane:
Interesting, I use `golint` directly from github (`go get -u golang.org/x/lint/golint`). It does not understand any `--version` parameter :-(. I wonder what `golang-1.10` image in ci/prow/verify uses.

Anyway, this looks harmless.

/lgtm
/approve

PR #52, opened 2023-09-07
Clean the repository, it will be reused for all CSI driver operators

jsafrane:
Clean everything, keep `make test` and `make verify` working, we will reuse these CI jobs shortly.

@openshift/storage


jsafrane:
/hold
for 4.15 to open

PR #53, opened 2023-09-08
STOR-1455: Add asset generator

jsafrane:
Add generator of CSI driver assets, with AWS EBS as the first one.

There is some documentation present in README.md and mainly in `godoc`. See the README for usage.

TODO: unit tests are missing, but I think I abstracted all dependencies away, so it should be straightforward to add them with minimal changes. I'm working on them.


jsafrane:
cc @openshift/storage

assets/patches/sidecars/node_driver_registrar.yaml
```diff
@@ -0,0 +1,28 @@
...
+    spec:
+      containers:
+        - name: csi-node-driver-registrar
+          securityContext:
+            privileged: true
+          image: ${NODE_DRIVER_REGISTRAR_IMAGE}
+          imagePullPolicy: IfNotPresent
+          args:
+            - --csi-address=/csi/csi.sock
+            - --kubelet-registration-path=/var/lib/kubelet/plugins/${DRIVER_NAME}/csi.sock
```
mpatlasov, line 12: I think `DRIVER_NAME` is too generic name for this specific purpose. Can we use `DRIVER_REG_SOCK_PATH` instead?

assets/patches/sidecars/node_driver_registrar.yaml
```diff
@@ -0,0 +1,28 @@
...
+            - --v=${LOG_LEVEL}
+          # Empty env. instead of nil for json patch to append new env. vars there
+          env: []
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "rm -rf /registration/${DRIVER_NAME}-reg.sock /csi/csi.sock"]
```
mpatlasov, line 19: This pattern doesn't work for some operators, e.g. IBM VPC and manila. We could use "rm -rf ${SOCK_FILES_TO_DELETE)" and then each operator would define its own specific list of files to delete.

assets/patches/sidecars/provisioner.yaml
```diff
@@ -0,0 +1,46 @@
...
+          # Empty env. instead of nil for json patch to append new env. vars there
+          env: []
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /var/lib/csi/sockets/pluginproxy/
+          resources:
+            requests:
+              memory: 50Mi
+              cpu: 10m
+        - name: provisioner-kube-rbac-proxy
```
mpatlasov, line 26: IBM VPC defines only `csi-provisioner` container, not `provisioner-kube-rbac-proxy` one. Should we make `provisioner-kube-rbac-proxy` optional?

assets/generated/README
```diff
@@ -0,0 +1 @@
+This directory contains yaml files generated by `make update`.
```
mpatlasov, line 1: New line missed.

assets/patches/sidecars/node_driver_registrar.yaml
jsafrane: This looks like a mistake in manila - node-driver-registrar should remove its own socket and not the driver's.

And [IBM VPC is odd](https://github.com/openshift/ibm-vpc-block-csi-driver-operator/blob/55390bc12acc933667dd9cf4e5459cc214cedb2a/assets/node.yaml#L70), `/registration/vpc.block.csi.ibm.io` should not exist.

assets/patches/sidecars/node_driver_registrar.yaml
jsafrane: But it is really the driver name. It's only used as part of two socket paths, so one `DRIVER_REG_SOCK_PATH` would not work either.  There is `/var/lib/kubelet/plugins/*/csi.sock` (driver's socket) and `/var/lib/kubelet/plugins_registry/*-reg.sock` (created by node-driver-registrar).

assets/patches/sidecars/node_driver_registrar.yaml
jsafrane: Btw, good catches for both of them! Thanks for reviewing all CSI drivers.

assets/patches/sidecars/provisioner.yaml
jsafrane: I would say all CSI drivers should export metrics, i.e. have provisioner-kube-rbac-proxy.

We need to sync with IBM, because they deploy the driver in their own managed OCP that's not HyperShift and I don't know how they collect metrics.

assets/generated/README
jsafrane: Fixed

assets/patches/sidecars/node_driver_registrar.yaml
mpatlasov: Yes, I agree (I missed that this name came from `assets/csidriver.yaml`). I still think that `CSI_DRIVER_NAME` would be better because it matches to `kind: CSIDriver` in csidriver.yaml, but now it's not so important -- it's up to you.

mpatlasov: Btw, if `DRIVER_NAME` is first-class citizen (defined in aws-ebs.go along with `AssetPrefix` and `AssetShortPrefix`), it's not nice that magic string `"ebs.csi.aws.com"` is duplicated many times in other yamls:
```
assets/base/node.yaml:41:            path: /var/lib/kubelet/plugins/ebs.csi.aws.com/
assets/drivers/aws-ebs/base/csidriver.yaml:4:  name: ebs.csi.aws.com
assets/drivers/aws-ebs/base/storageclass_gp2.yaml:8:provisioner: ebs.csi.aws.com
assets/drivers/aws-ebs/base/storageclass_gp3.yaml:10:provisioner: ebs.csi.aws.com
assets/drivers/aws-ebs/base/volumesnapshotclass.yaml:7:driver: ebs.csi.aws.com
```
Can we use `${DRIVER_NAME}` everywhere instead of `"ebs.csi.aws.com"` ?

pkg/generated-assets/sanitize.go
```diff
@@ -0,0 +1,25 @@
...
+	sigyaml.Unmarshal(src, &obj)
+	bytes, err := sigyaml.Marshal(obj)
+	if err != nil {
+		return nil, err
+	}
+	return bytes, nil
+}
+
+// MustSanitize reorders fields in a YAML file in a canonical order, so they can be compared easily with `diff`.
+func MustSanitize(src []byte) []byte {
```
mpatlasov, line 19: Not used anywhere

pkg/generator/asset_generator.go
```diff
@@ -0,0 +1,363 @@
...
+	return nil
+}
+
+// Apply all controller patches in the generator config (CSIDriverGeneratorConfig.ControllerConfig.AssetPatches)
+func (gen *AssetGenerator) patchController() error {
+	for _, patch := range gen.operatorConfig.ControllerConfig.AssetPatches {
+		if !patch.ClusterFlavours.Has(gen.flavour) {
+			continue
+		}
+		switch patch.GeneratedAssetName {
```
mpatlasov, line 81: The switch looks rather clumsy. I think it's worthy to have full mapping like `gen.generatedAssets.ControllerAssets[]` to have `generated_assets.ControllerDeploymentAssetName` included. Then `assetYAML := gen.generatedAssets.ControllerAssets[patch.GeneratedAssetName]` would work unconditionally.

And `gen.generatedAssets.ControllerStaticResources[]` could be a slice of that full mapping.

pkg/generator/asset_generator.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+// Apply all patches in the generator config (CSIDriverGeneratorConfig.GuestConfig.AssetPatches)
+func (gen *AssetGenerator) patchGuest() error {
+	// Patch everything, including the CSI driver DaemonSet.
+	for _, patch := range gen.operatorConfig.GuestConfig.AssetPatches {
+		if !patch.ClusterFlavours.Has(gen.flavour) {
+			continue
+		}
+		switch patch.GeneratedAssetName {
```
mpatlasov, line 287: Same as for patchController(): let's have all mappings in a single map like `gen.generatedAssets.GuestAssests[]` and specific mappings like `gen.generatedAssets.GuestStorageClassAssets[]` are slices there. Btw, this switch is hard to review because `gen.operatorConfig.GuestConfig.AssetPatches` is empty. Was it intended?

pkg/generated-assets/sanitize.go
jsafrane: Removed

pkg/generator/asset_generator.go
jsafrane: +1 for clumsiness. I can add controller Deployment + node DaemonSet directly to Controller / NodeAssets easily, but it gets complicated with StorageClasses / VolumeSnapshotClasses below.

I added a separate patch moving everything just to two lists, `ControllerAssets` + `GuestAssets`, and detecting their `kind:` in runtime. This detection involves parsing all YAML files, possibly multiple times, but it should be used only during operator startup.

pkg/generator/asset_generator.go
jsafrane: See above

gnufied:
I am still reading it and learning but one quick feedback IMO is, we perhaps need more hierarchy that mimicks kustomize more closely.

For example to really learn where the patches in top level assets folder apply, we will have to dig through the code.  kustomize is used for "composing" patches, but reading the assets directory, it is not immediately obvious how the patches are composed.  Such as, For generating `deployment.yaml` on hypershift - I can't easily tell in the order various patches are applied.

jsafrane:
> it is not immediately obvious how the patches are composed

That's result of more reusable composition. Look at this sidecar definition:

```
			Sidecars: []generator.SidecarConfig{
				common.DefaultProvisionerWithSnapshots.WithExtraArguments(
					"--default-fstype=ext4",
					"--feature-gates=Topology=true",
					"--extra-create-metadata=true",
					"--timeout=60s",
				),
```
It actually tells the generator:
* to apply patches/sidecars/provisioner.yaml to controller.yaml
  * with extra arguments added to cmdline via go code, it would be ugly with kustomize
  * with patches/hypershift/sidecar_kubeconfig.yaml.patch applied on HyperShift
* with base/rbac/main_provisioner_binding.yaml and base/rbac/volumesnapshot_reader_provisioner_binding.yaml added to static assets.

And that's just 1 sidecar.

I agree it's not obvious how are the final assets generated. But if I expand all this knowledge into pkg/config/aws-ebs/aws-ebs.go, it would be IMO quite ugly mess there. And *all* this would be copied into all CSI drivers, with minor and confusing changes e.g. if a driver does / does not support snapshots (to get the right RBAC).

Let me know if you have a way how to make the input as readable and obvious as possible and without copying large chunks of code around for various drivers.

I could add more logs to the generator or even create a separate changelog file for each generated asset how it got created. I am not sure it would help much though.

jsafrane:
Maybe another idea: the generator could generate kustomize file. I think it would be too meta - a generator producing input for another generator.

assets/patches/sidecars/node_driver_registrar.yaml
jsafrane: I fixed `assets/base/node.yaml`, because it's the base of all daemonsets and the driver name must be replaced there.

Files in `assets/drivers/aws-ebs` are already aws-ebs specific, I think `${DRIVER_NAME}` does not make them better or more readable.

mpatlasov:
> Maybe another idea: the generator could generate kustomize file. I think it would be too meta ...

Agreed. meta-on-meta would'n make it easier to understand or cleaner implementation.

jsafrane:
Move assets around
    
- Add a prefix to most patches about what they patch - like controller_
      patches a Deployment, node_ patches a DaemonSet, service_ patches
      Service.
- Add a comment to most patches about what they do.
- Removed controller_proxy.patch, it's not used.
- Move patches/ to common/
- Move drivers/ to overlays/
- Move aws-ebs.go to assets/overlays/aws-ebs, so it's close to yaml files.
- Move generated files to overlays/aws-ebs/generated, so it's close to the source .go and source yaml files.

assets/common/sidecars/kube_rbac_proxy.yaml
```diff
@@ -0,0 +1,27 @@
...
+          - containerPort: ${EXPOSED_METRICS_PORT}
+            name: ${PORT_NAME}
+            protocol: TCP
+          resources:
+            requests:
+              memory: 20Mi
+              cpu: 10m
+          volumeMounts:
+          - mountPath: /etc/tls/private
+            name: metrics-serving-cert
```
gnufied, line 27: So this particular kube-rbac-proxy is for CSI driver itself. For other containers the rbac proxy sidecar is named accordingly but this one sticks out a bit. May be we should change it?

assets/overlays/aws-ebs/aws_ebs.go
```diff
@@ -0,0 +1,72 @@
...
+	return &generator.CSIDriverGeneratorConfig{
+		AssetPrefix:      "aws-ebs-csi-driver",
+		AssetShortPrefix: "ebs",
+		DriverName:       "ebs.csi.aws.com",
+
+		ControllerConfig: &generator.ControlPlaneConfig{
+			DeploymentTemplateAssetName: "overlays/aws-ebs/patches/controller_add_driver.yaml",
+			LivenessProbePort:           10301,
+			MetricsPorts: []generator.MetricsPort{
+				{
```
gnufied, line 19: Why is this an array? Configuration here I feel defines configuration of the driver. Can we have more than one of these in some cases?

assets/overlays/aws-ebs/aws_ebs.go
```diff
@@ -0,0 +1,72 @@
...
+					"--timeout=300s",
+					"--extra-create-metadata",
+				),
+				common.DefaultLivenessProbe.WithExtraArguments(
+					"--probe-timeout=3s",
+				),
+			},
+			Assets: common.DefaultControllerAssets,
+			AssetPatches: common.DefaultAssetPatches.WithPatches(generator.HyperShiftOnly,
+				"controller.yaml", "overlays/aws-ebs/patches/controller_add_hypershift_controller_minter.yaml",
```
gnufied, line 51: It is bit weird to see use of sets here. So this basically defines - in hypershift environment, for `controller.yaml` apply this patch?

gnufied: I feel like there is an assumption that `ControlPlaneConfig` will produce `controller.yaml`. Can we declare this as a field, which defaults to `controller.yaml` so as user does not have specify it, but it can be used to refer the code and verify what resource this section produces.

pkg/generated-assets/sanitize.go
```diff
@@ -1,16 +1,43 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if line[0] != '#' {
+			return comments.Bytes(), nil
+		}
+		comments.Write(line)
+	}
+
+	return nil, nil
```
mpatlasov, line 42: `go vet` is unhappy because this line is unreachable. I'd suggest:
```
- return nil, nil
- // We can never be here.
```
I'd also appreciate a comment explaining your attitude toward asset files with comments only (w/o actual content). Something short and simple like this would work for me:
```
// We don't expect asset yaml files containing only comments. initialComments() would fail in such cases intentionally.
```

pkg/generated-assets/generated_assets.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		if err := os.WriteFile(path, data, 0644); err != nil {
+			return fmt.Errorf("failed to write asset %s: %w", name, err)
+		}
+	}
+	return nil
+}
+
+// NewFromAssets loads the assets from the given directory.
+func NewFromAssets(reader resourceapply.AssetFunc, dir string) (*CSIDriverAssets, error) {
```
mpatlasov, line 220: I couldn't figure our who and when calls `NewFromAssets()`. Ideas?
(it's worthy to have at least one caller, because strictly speaking its validity depends on `reader` arg -- hard to guess if it's simply `os.ReadFile()` or something more sophisticated)

pkg/generated-assets/generated_assets.go
```diff
@@ -0,0 +1,272 @@
...
+	manifestFileName              = "manifests.yaml"
+	ControllerDeploymentAssetName = "controller.yaml"
+	NodeDaemonSetAssetName        = "node.yaml"
+	MetricServiceAssetName        = "service.yaml"
+	MetricServiceMonitorAssetName = "servicemonitor.yaml"
+)
+
+// CSIDriverAssets contains all the assets required to deploy the CSI driver.
+// The assets are stored as raw bytes, and can be retrieved using GetAsset.
+// The assets can be saved to a directory using Save and loaded using NewFromAssets function.
```
mpatlasov, line 41: The comment doesn't help to understand who and when is supposed to call NewFromAssets function.

pkg/generator/patch.go
```diff
@@ -0,0 +1,144 @@
...
+	return gen.applyStrategicMergePatch(sourceYAML, patchAssetName, extraReplacements)
+}
+
+// Apply strategic merge patch to the source YAML.
+func (gen *AssetGenerator) applyStrategicMergePatch(sourceYAML []byte, patchAssetName string, extraReplacements []string) ([]byte, error) {
+	patchYAML := gen.mustReadAsset(patchAssetName, extraReplacements)
+	opts := yaml.MergeOptions{ListIncreaseDirection: yaml.MergeOptionsListAppend}
+	ret, err := merge2.MergeStrings(string(patchYAML), string(sourceYAML), false, opts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to apply aseet %s: %v", patchAssetName, err)
```
mpatlasov, line 47: `%v` looks as a typo. Should we use `%w` everywhere, or there is a reason to prefer `%v` here?

mpatlasov: Also: `s/aseet/asset` :)

pkg/generator/patch.go
```diff
@@ -0,0 +1,144 @@
...
+			return nil, err
+		}
+	}
+
+	// Apply the sidecar as strategic merge patch against the source YAML. This add the sidecar as a new container
+	// into the source deployment YAML.
+	opts := yaml.MergeOptions{ListIncreaseDirection: yaml.MergeOptionsListAppend}
+	ret, err := merge2.MergeStrings(string(sidecarYAML), string(sourceYAML), false, opts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to apply asset %s: %v", sidecarAssetName, err)
```
mpatlasov: `%v` looks as a typo. Should we use `%w` everywhere, or there is a reason to prefer `%v` here?

mpatlasov:
@jsafrane , the PR currently consists of 9 commits. The PR will be merged as soon as you fix `ci/prow/verify`. I don't see much value in keeping all 9 commits in the history forever. You could squeeze them into single commit (or in a fewer number of commits). I'm rather neutral about it, so take it just as head-up.

gnufied:
This PR is so large that, firefox/chrome hangs while trying to review. Overall I am happy with the shape of it. There are some minor nits about various implementation details but IMO those can be fixed later on (and hence my lgtm).

jsafrane:
/lgtm cancel
I want to fix the review comments first + clean up the yaml logs

assets/overlays/aws-ebs/aws_ebs.go
jsafrane: At least vSphere CSI driver has two containers and each of them can have its own metric port. We don't scrape syncer metrics, but perhaps we should.

assets/common/sidecars/kube_rbac_proxy.yaml
jsafrane: Renamed to driver_kube_rbac_proxy.yaml + added a comment to the yaml file.

assets/overlays/aws-ebs/aws_ebs.go
jsafrane: Yes, exactly. I am not sure there is a patch that's applied to both flavours, but there are _assets_ that are (e.g. storageclass_gp2.yaml in this file) and then there are flavour-specific patches (`common/standalone/controller_add_affinity.yaml` / `common/hypershift/controller_add_affinity_tolerations.yaml`).

pkg/generated-assets/generated_assets.go
jsafrane: It will be used by a subsequent PR to actually load the files. See https://github.com/openshift/aws-ebs-csi-driver-operator/pull/258/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6 and expand starter.go

pkg/generated-assets/generated_assets.go
jsafrane: I added few comments around, here, to `Save()` and to `NewFromAssets()`

pkg/generator/patch.go
jsafrane: Fixed.

pkg/generator/patch.go
jsafrane: Fixed.

pkg/generated-assets/sanitize.go
jsafrane: Good point about files containing only comments. I added a separate error for that.
(+ there should be an unit test for this...)

jsafrane:
I squashed all commits except the yaml log, I'm fixing it right now.

> This PR is so large that, firefox/chrome hangs while trying to review

I blame >4000 files in vendor. There is a separate commit with it.

jsafrane:
I reworked the yaml history. I moved it to a separate file, `yaml.go` and put `ApplyStrategicMergePatch` + `ApplyJSONPatch there`, so it can apply a patch + record it in the history.

And I polished the output a bit.

jsafrane:
Rebased, with new `Makefile` targets to test & verify both the legacy operator(s) and the new one.

cmd/generator/main.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/openshift/csi-operator/pkg/generator"
+)
+
+// generator is a tool that generates assets for CSI driver operators.
+// It is intended to be used *before* building the operators, using `make update`.
+//
+// The generated assets will be then compiled into the operator binaries using assets.go.
+func main() {
+	// TODO: AWS EBS is hardcoded, add support for other CSI driver operators
+	flavour := flag.String("flavour", "standalone", "cluster flavour")
```
bertinatto, line 17: `flavour` -> british English detected :joy: 

nit: add possible flavors to usage

gnufied:
lgtm

bertinatto (commented):
/lgtm
We agreed to merge this and follow-up in subsequent PRs.

PR #55, opened 2023-09-21
Merge aws ebs driver operator

gnufied:
Using git-subtree for merging aws-ebs-driver-operator

PR #56, opened 2023-09-26
Add aws ebs operator legacy

gnufied:
Keep the same repository name for the image.


jsafrane:
For adoption in CI we will need also `.ci-operator.yaml` in the top level dir. It can be added as as subsequent PR though.

And the initial commit message still does not contain source of the commit as @mpatlasov suggested in https://github.com/openshift/csi-operator/pull/55#issuecomment-1730668191. I'm not sure how easy / hard it is to add it there.

gnufied:
Added ci-operator file and update merge commit to reflect sha that was merged. I couldn't edit the squash commit generated by subtree.

mpatlasov:
> I couldn't edit the squash commit generated by subtree.

`c0f610b` tells clearly what was the source (`openshift/aws-ebs-csi-driver-operator`). I'm fine with it -- no need to wrestle with squash commit, imo.

PR #60, opened 2023-10-25
Move few things around

jsafrane:
1. Move `pkg/config/common`, with definition of common sidecar parameters, to the `pkg/generator/common`. `pkg/config` is now empty and removed. Originally, `config` was the place that had aws-ebs generator + operator config, but it's not _only_ config, it's lot of webhooks now. See the point below.

2. Move AWS EBS code from `assets` to `pkg/driver/aws-ebs`. It's not an asset, it's lot of webhook code + operator + generator config. So let it be in `pkg/driver/aws-ebs`, `pkg/driver/azure-disk` etc.

And when at it, I made a `const` with the AWS EBS asset base path. Just in case we want to move it in future. All paths in the AWS EBS generator config were hardcoded, but path to the generated assets had a const. Now all paths are based on a const.


jsafrane:
/close
in favor of https://github.com/openshift/csi-operator/pull/61

PR #62, opened 2023-10-25
Verify generated assets in `make verify`

jsafrane:
Add `verify-generated-assets` make target that just re-generates the assets + calls `git diff`. It will produce false negatives when users have unstaged changes in their workdir, but it's good enough for CI. It's similar to what [openshift/api uses](https://github.com/openshift/api/blob/b8a18fdc040de6208d49111ac776ac08b98e6bb7/Makefile#L67).


jsafrane:
cc @openshift/storage

PR #64, opened 2023-10-26
Drop Guest prefix from client names

jsafrane:
Most of the clients are for the guest cluster, keep ControlPlane for the mgmt cluster.


jsafrane:
cc @openshift/storage

PR #59, opened 2023-10-16
OCPBUGS-21593: CVE-2023-44487: bump golang.org/x/net to v0.17.0

dobsonj:
cc @openshift/storage

PR #68, opened 2023-11-13
Fix vendoring of golang.org/x/net

jsafrane:
Fix verify-deps by adding missing files added by `go mod tidy` && `go mod vendor`. Not sure how the current vendor got merged without them.

@openshift/storage


dobsonj:
I don't understand how this happened, but https://github.com/openshift/csi-operator/pull/59 actually _removed_ those files. The verify-deps job did not exist when that PR merged though. Strange.

/lgtm

PR #72, opened 2023-11-21
STOR-1400: Sync sidecar arguments with upstream

jsafrane:
Update the operator to use the same sidecar arguments (timeouts, QPS, worker threads) as upstream.

See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1824 and https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1834.

Note that this partially reverts https://github.com/openshift/aws-ebs-csi-driver-operator/pull/248, external-resizer + snapshotter will get `60s` timeout instead of `300s` that we have in OCP.

cc @openshift/storage


jsafrane:
Updated external-resizer, it uses `-workers=` instead of `-worker-threads` like the other sidecars.

jsafrane:
/retest 
snapshot tests timed out + hypershift could not install its mgmt cluster due to image registry operator.

PR #73, opened 2023-11-21
STOR-1722:  Add support hypershift for azure-disk

gnufied:
[STOR-1722](https://issues.redhat.com//browse/STOR-1722)


assets/overlays/azure-disk/patches/controller_add_driver.yaml
```diff
@@ -0,0 +1,89 @@
...
+apiVersion: apps/v1
+metadata:
+  annotations:
+    config.openshift.io/inject-proxy: csi-driver
+    config.openshift.io/inject-proxy-cabundle: csi-driver
+spec:
+  template:
+    spec:
+      initContainers:
+        # Merge /etc/kubernetes/cloud.conf (on the host) with secret "azure-disk-credentials" into "merged-cloud-config" emptydir.
```
jsafrane, line 11: This is for another PR & discussion, but this will not work on HyperShift at all. /etc/kubernetes/cloud.conf on the host has no relation to the Azure config + credentials that should be used to manage storage of a guest cluster.  For example, the mgmt cluster does not need to be on Azure at all.
We will need to figure out something else.

pkg/clients/builder.go
```diff
@@ -84,6 +84,7 @@ func (b *Builder) BuildOrDie(ctx context.Context) *Clients {
 		var err error
 		guestKubeConfig, err = client.GetKubeConfigOrInClusterConfig(b.guestKubeConfigFile, nil)
 		if err != nil {
+			klog.Errorf("error reading guesKubeConfig from %s: %v", b.guestKubeConfigFile, err)
 			panic(err)
```
jsafrane, line 88: Would be `klog.Fatalf` + no panic better? I don't know how the actual output looks like here. This applies to all panics below.

pkg/driver/azure-disk/azure_disk.go
```diff
@@ -0,0 +1,376 @@
...
+
+	infra, err := c.ConfigClientSet.ConfigV1().Infrastructures().Get(ctx, infrastructureName, metaV1.GetOptions{})
+	if err != nil {
+		klog.Errorf("unable to fetch infra object from guest cluster with: %v", err)
+		return nil, fmt.Errorf("unable to fetch infra object: %v", err)
+	}
+
+	insideStackHub := runningInAzureStackHub(infra)
+
+	// We need featuregate accessor made available to the operator pods
```
jsafrane, line 148: I think it would make sense to move the feature gate code to starter.go / a separate file, IMO we can expect that more operators will use it. `client.Clients` can have a field (accessor?) to read the gates.

pkg/operator/config/config.go
```diff
@@ -50,6 +55,12 @@ type OperatorControllerConfig struct {
 	// List of hooks should be run on the storage classes.
 	// No informers here, because StorageClassController does not accept any.
 	StorageClassHooks []csistorageclasscontroller.StorageClassHookFunc
+
+	// ExtraReplacements defines additional replacements that should be made to assets
+	ExtraReplacementsFunc func() []string
+
+	// SnapshotAssetName provides a way to override snapshot asset name that will be used
+	SnapshotAssetNameFunc func() []string
```
jsafrane, line 63: I think VolumeSnapshotClasses and StorageClasses should have the same handling. I.e. there should be  `VolumeSnapshotClassHooks` that would set ASH specific parameters and not a separate yaml file. I know, it implies a whole new controller. Maybe it would be better in the long run than the StaticResourceController that we have now.

pkg/driver/azure-disk/azure_disk.go
gnufied: +1 , I can do that.

assets/overlays/azure-disk/patches/controller_add_driver.yaml
gnufied: Yeah this doesn't work on hypershift. This is one of the last blocking items. This prevents rollout of deployment in hypershift. We will have to figure something out.

gnufied: This is now fixed and tested with hypershift and it works as expected. I ended up using a certain secret in hypershift envionment.

pkg/clients/builder.go
gnufied: fixed

pkg/operator/config/config.go
gnufied: Isn't less code better? :-) 

Seriously though - if all storageclass controller does is, performs cosmetic changes to storageclass, then it is perhaps a mistake to carry 300 lines of code (and tests), just for tweaking an parameter. Not to mention, a whole new controller results in additional api-usage, events, conditions and what not.

pkg/operator/config/config.go
```diff
@@ -50,6 +55,12 @@ type OperatorControllerConfig struct {
 	// List of hooks should be run on the storage classes.
 	// No informers here, because StorageClassController does not accept any.
 	StorageClassHooks []csistorageclasscontroller.StorageClassHookFunc
+
+	// ExtraReplacements defines additional replacements that should be made to assets
+	ExtraReplacementsFunc func() []string
```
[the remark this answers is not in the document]
gnufied: It could go either way, I am happy to change it.

pkg/operator/starter.go
```diff
@@ -129,15 +152,23 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		)
 	}
 
+	var snapshotAssetNames []string
+
+	if csiOperatorControllerConfig.SnapshotAssetNameFunc != nil {
+		snapshotAssetNames = csiOperatorControllerConfig.SnapshotAssetNameFunc()
+	} else {
+		snapshotAssetNames = a.GetVolumeSnapshotClassAssetNames()
+	}
```
dobsonj, line 161: Instead of adding this if-then-else: could we ensure SnapshotAssetNameFunc is never nil and call it unconditionally? Then the aws operator config for example would just initialize `SnapshotAssetNameFunc = a.GetVolumeSnapshotClassAssetNames()`

pkg/operator/starter.go
```diff
@@ -38,19 +39,30 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
 	assetDir := filepath.Join(opConfig.AssetDir, string(flavour))
 	a, err := generated_assets.NewFromAssets(assets.ReadFile, assetDir)
 	if err != nil {
 		return err
 	}
-	a.SetReplacements(operator.DefaultReplacements(controlPlaneNamespace))
+	defaultReplacements := operator.DefaultReplacements(controlPlaneNamespace)
+	if csiOperatorControllerConfig.ExtraReplacementsFunc != nil {
+		defaultReplacements = append(defaultReplacements, csiOperatorControllerConfig.ExtraReplacementsFunc()...)
+	}
```
dobsonj, line 63: Similar idea here, could we require each platform to define a replacement func? It could just return the value of DefaultReplacements() if there is nothing more to add.

pkg/driver/azure-disk/azure_disk.go
```diff
@@ -0,0 +1,411 @@
...
+		value := fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/diskEncryptionSets/%s", encset.SubscriptionID, encset.ResourceGroup, encset.Name)
+		klog.V(4).Infof("Setting %s = %s in StorageClass %s", diskEncryptionSetID, value, class.Name)
+		class.Parameters[diskEncryptionSetID] = value
+		return nil
+	}
+
+	// Explicitly instantiate ClusterCSIDriver informer, so it is synced during WaitForCacheSync
+	// and thus its lister is populated
+	// _ = c.OperatorInformers.Operator().V1().ClusterCSIDrivers().Informer()
+	return hook
```
dobsonj, line 388: Do you think this is a bug in the current code? Should we have a bug to track it, since it probably affects other operators too?

Dockerfile.azure-disk
```diff
@@ -0,0 +1,12 @@
+# Temporary dockerfile for the new Azure Disk CSI driver operator. To be moved into Dockerfile.aws-ebs
+# soon
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.15 AS builder
```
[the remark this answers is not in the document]
dobsonj: Actually azure-disk-csi-driver-operator is still rhel-8 at the moment:
https://github.com/openshift/azure-disk-csi-driver-operator/blob/master/Dockerfile.openshift#L1
but I'd expect it to get updated soon?
https://github.com/openshift-eng/ocp-build-data/blob/openshift-4.15/images/local-storage-diskmaker.yml#L12

assets/overlays/azure-disk/generated/standalone/controller.yaml
```diff
@@ -0,0 +1,372 @@
...
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+        volumeMounts:
+        - mountPath: /var/lib/csi/sockets/pluginproxy/
+          name: socket-dir
+        - mountPath: /etc/kubernetes/
+          name: cloud-config
+          readOnly: true
```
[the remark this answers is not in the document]
gnufied: So I checked these mount points on host and inside the driver pod and they are empty. I am thinking that, they were copied verbatim from upstream. But Microsoft actually uses these via MSI tokens but our deployment is not yet setup to use MSI tokens and hence we should revisit this when we are actually ready to support msi tokens.

pkg/driver/azure-disk/azure_disk.go
gnufied: It is not a bug as such. @jsafrane used this trick to get informers to populated but I don't actually need this workaround because I explicitly added objects to informers before running tests and it worked just fine.  I can do the same aws-ebs operator and we won't need this anymore.

pkg/driver/azure-disk/azure_disk.go
```diff
@@ -0,0 +1,411 @@
...
+				},
+			},
+			SidecarLocalMetricsPortStart:   commongenerator.AzureDiskLoopbackMetricsPortStart + 1,
+			SidecarExposedMetricsPortStart: commongenerator.AzureDiskExposedMetricsPortStart + 1,
+			Sidecars: []generator.SidecarConfig{
+				commongenerator.DefaultProvisionerWithSnapshots.WithExtraArguments(
+					"--default-fstype=ext4",
+					"--feature-gates=Topology=true",
+					"--extra-create-metadata=true",
+					"--timeout=60s",
```
jsafrane, line 85: This is `30s` in the original openshift/azure-disk-csi-driver-operator. And there are some parameters missing
```
        - --http-endpoint=localhost:8202
        - --strict-topology=true
        - --worker-threads=100
        - --kube-api-qps=50
        - --kube-api-burst=100
```

assets/overlays/azure-disk/base/csidriver.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: storage.k8s.io/v1
+kind: CSIDriver
+metadata:
+  name: disk.csi.azure.com
+  annotations:
+    # Used to determine if a CSI driver was created by OCP or by 3rd party operator / helm / yaml files.
+    csi.openshift.io/managed: "true"
+spec:
```
jsafrane, line 8: ` seLinuxMount: true`  is missing

pkg/driver/azure-disk/azure_disk.go
```diff
@@ -0,0 +1,411 @@
...
+				),
+				commongenerator.DefaultAttacher.WithExtraArguments(
+					"--timeout=60s",
```
jsafrane, line 88:
```
        - --timeout=600s
        - --worker-threads=500
        - --kube-api-qps=50
        - --kube-api-burst=100
```

pkg/driver/azure-disk/azure_disk.go
```diff
@@ -0,0 +1,411 @@
...
+				),
+				commongenerator.DefaultResizer.WithExtraArguments(
+					"--timeout=300s",
```
jsafrane, line 91:
```
        - --timeout=240s
       - --handle-volume-inuse-error=true
```

pkg/driver/azure-disk/azure_disk.go
```diff
@@ -0,0 +1,411 @@
...
+				),
+				commongenerator.DefaultSnapshotter.WithExtraArguments(
+					"--timeout=300s",
```
jsafrane, line 94: This is probably an oversight, but the original operator did not set any timeout.
Upstream has `            - "--timeout=1200s"`

pkg/driver/azure-disk/azure_disk.go
```diff
@@ -0,0 +1,411 @@
...
+					"--extra-create-metadata",
```
jsafrane, line 95: neither the old operator or upstream have `- --extra-create-metadata`

jsafrane (commented):
I reviewed just the yaml files here. I still need to check volumes in the CSI driver deployment / daemonset, they're quite different to the original ones.

assets/overlays/azure-disk/generated/standalone/node.yaml
```diff
@@ -0,0 +1,220 @@
...
+        - mountPath: /dev
+          name: device-dir
+        - mountPath: /sys/bus/scsi/devices
+          name: sys-devices-dir
+        - mountPath: /sys/class/scsi_host/
+          name: scsi-host-dir
+        - mountPath: /etc/selinux
+          name: etc-selinux
+        - mountPath: /sys/fs
+          name: sys-fs
```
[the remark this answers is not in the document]
gnufied: same as above.

pkg/clients/builder.go
```diff
@@ -47,14 +47,18 @@ func NewBuilder(userAgent string, csiDriverName string, controllerConfig *contro
...
 }
 
 // WithHyperShiftGuest sets the kubeconfig file for the guest cluster. Usable only on HyperShift.
-func (b *Builder) WithHyperShiftGuest(kubeConfigFile string) *Builder {
+func (b *Builder) WithHyperShiftGuest(kubeConfigFile string, cloudConfigNamespace string) *Builder {
 	b.guestKubeConfigFile = kubeConfigFile
+	if cloudConfigNamespace != "" {
+		b.guestNamespaces = append(b.guestNamespaces, cloudConfigNamespace)
+	} else {
+		b.guestNamespaces = append(b.guestNamespaces, ManagedConfigNamespace)
```
jsafrane, line 60: I think we can remove the whole `ManagedConfigNamespace` variable and AWS EBS can use `CloudConfigNamespace: cloudConfigNamespace` (where cloudConfigNamespace is [this](https://github.com/openshift/csi-operator/blob/323d0d0032a21637ba98cd4bac0ef932bf147fe0/pkg/driver/aws-ebs/aws_ebs.go#L33))

pkg/operator/config/config.go
jsafrane: We may need the same defaulting logic that we have in StorageClassController in volume snapshot too. My point is, storage classes and snapshot classes are +/- the same objects and thus should have the same behavior (they don't have that now) and code.
(Would there be a way to have a common ClassController?)

assets/overlays/azure-disk/generated/standalone/controller.yaml
jsafrane, line 97: I think we still need:

```
- mountPath: /var/run/secrets/openshift/serviceaccount
          name: bound-sa-token
```
 + entry in `volumes` section. We support managed identities on [standalone cluster](https://docs.openshift.com/container-platform/4.14/installing/installing_azure/installing-azure-private.html#installing-azure-with-short-term-creds_installing-azure-private) too.

And please ask QE to test it explicitly (or test it by yourself). This can easily break.

assets/overlays/azure-disk/generated/standalone/node.yaml
```diff
@@ -0,0 +1,220 @@
...
+        name: cloud-config
+      - hostPath:
+          path: /sys/bus/scsi/devices
+          type: Directory
+        name: sys-devices-dir
+      - hostPath:
+          path: /sys/class/scsi_host/
+          type: Directory
+        name: scsi-host-dir
+      - name: merged-cloud-config
```
jsafrane, line 216: I did not check where this comes from, but empty `volume` is weird.

pkg/operator/config/config.go
jsafrane: Ok, this is actually the biggest issue in this PR (I mean the most complicated to fix). I'm fine with a TODO comment and let's fix it later.

pkg/operator/config/config.go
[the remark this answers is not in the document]
jsafrane: Assuming there are feature gates available in `client.Clients`, this can be a static array populated in `GetAzureDiskOperatorControllerConfig`. The I guess it's file to have just `[]string` here, it may get handy for other CSI drivers.

gnufied:
(failing external hypershift test looks unrelated to this change)

/retest

pkg/driver/azure-disk/azure_disk.go
jsafrane: This has not been addressed yet.

assets/overlays/azure-disk/generated/standalone/node.yaml
jsafrane: This has not been addressed yet. I think we just need:
```
- name: merged-cloud-config
   emptyDir: {}
```

pkg/clients/builder.go
```diff
@@ -97,40 +105,55 @@ func (b *Builder) BuildOrDie(ctx context.Context) *Clients {
...
+}
 
-	clients.ConfigClientSet = cfgclientset.NewForConfigOrDie(guestKubeConfig)
-	clients.ConfigInformers = cfginformers.NewSharedInformerFactory(clients.ConfigClientSet, b.resync)
+func (b *Builder) GetClient() *Clients {
+	return b.client
+}
 
-	return clients
+func (b *Builder) AddSnpshotClient(ctx context.Context) (snapshotclientset.Interface, error) {
```
jsafrane, line 152: typo: AddSnpshotClient

pkg/clients/builder.go
jsafrane: What about this comment? It's easy and it would make the common code a bit easier to read.

gnufied:
The stackhub CSI is failing for unrelated reason - 

```
sion=2015-01-01 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f9bb28cd430>: Failed to establish a new connection: [Errno -2] Name or service not known'))
```

But I have tested the integration.

ropatil010:
Added the test results here: https://issues.redhat.com/browse/STOR-1707
/test e2e-azurestack-csi
/label qe-approved

PR #84, opened 2023-12-12
OCPBUGS-25125: Create RBAC objects first

bertinatto:
Make sure SAs and Roles are created before bindings, followed by other resources like DS and Deployments.

### DIFF
--- a/pkg/generated-assets/generated_assets.go
+++ b/pkg/generated-assets/generated_assets.go
@@ -84,7 +84,7 @@ func (a *CSIDriverAssets) getRawAsset(generatedAssetName string) ([]byte, error)
 // namespace or standalone cluster. These assets should be managed by a StaticResourcesController.
 // Any Deployment is filtered out from the list, they're supposed to be handled by DeploymentController.
 func (a *CSIDriverAssets) GetControllerStaticAssetNames() []string {
-	var names []string
+	assets := new(AssetOrderer)
 	for name, yaml := range a.ControllerAssets {
 		kind, err := getYAMLKind(yaml)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


bertinatto:
/hold
Just testing this approach.

pkg/generated-assets/generated_assets.go
```diff
@@ -275,3 +275,22 @@ func getYAMLKind(yaml []byte) (string, error) {
 	}
 	return gvk.GroupKind().String(), nil
 }
+
+func sortAssets(names []string) []string {
+	var stage1, stage2, stage3 []string
+	for i := range names {
+		switch {
+		case strings.HasSuffix(names[i], "_sa.yaml") || strings.HasSuffix(names[i], "_role.yaml"):
```
jsafrane, line 283: I think service accounts should be created as the last, after all roles and rolebindings are created. A missing SA will produce a different error message than "unable to validate against any security context constraint" and the test will more likely pass ;-)

pkg/generated-assets/generated_assets.go
```diff
@@ -275,3 +275,22 @@ func getYAMLKind(yaml []byte) (string, error) {
...
+			stage1 = append(stage1, names[i])
+		case strings.HasSuffix(names[i], "_binding.yaml"):
```
jsafrane, line 285: I don't like depending on specific file names. It would be more robust to sort by the yaml content. `GetControllerStaticAssetNames` already parses the yaml files and has their `kind`, it should be possible to sort by it. Something like:

```
func getCreationOrder(kind string) int {
    if rbacKinds.Has(kind) {
        return  0
    }
    return 1
}
```

jsafrane:
I like this PR! Only if CI worked...

jsafrane:
I can see in [e2e logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-operator/84/pull-ci-openshift-csi-operator-master-e2e-aws-csi/1735334787519877120/artifacts/e2e-aws-csi/gather-extra/artifacts/pods/openshift-cluster-csi-drivers_aws-ebs-csi-driver-operator-5589d68bb4-6mh66_aws-ebs-csi-driver-operator.log) that both StaticResourceControllers create RBAC objects first and then ServiceAccounts / Services. 

/lgtm

PR #89, opened 2023-12-18
OCPBUGS-24245: Add selinux

jsafrane:
Restore `seLinuxMount: true` in AWS EBS CSIDriver instance that we lost when moving to csi-operator repo.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #117, opened 2024-01-12
OCPBUGS-27056: Update to RHEL9 and go 1.21

jsafrane:
This is required by ART. [ART bot](https://github.com/openshift/csi-operator/pull/) is just not able to regenerate go.sum with the new go.


bertinatto (commented):
/lgtm

PR #156, opened 2024-02-02
STOR-1574: Bump all deps

jsafrane:
Updating all dependencies using `go get -u ./...`.
/label px-approved
/label docs-approved


jsafrane:
Rebased

PR #181, opened 2024-02-28
Revert "STOR-1762: Add azure file hypershift"

gnufied:
Reverts openshift/csi-operator#163


jsafrane:
Why is AzureFile being reverted? I think everything works (nightlies are green), at least on standalone OCP. We haven't received any complaints about HyperShift.

PR #199, opened 2024-03-12
OCPBUGS-23260: Explicitly reserve 1 attachment for the root disk

jsafrane:
Using a newly introduced option --reserved-volume-attachments, explicitly reserve 1 AWS EBS volume attachment for the root disk and leave the other attachment to PVs/PVCs.

This disables bad heuristics in the CSI driver, see the linked bug.

/hold
for https://github.com/openshift/aws-ebs-csi-driver/pull/260 to be merged first


bertinatto (commented):
/lgtm

jsafrane:
/label acknowledge-critical-fixes-only
This is an important bugfix

PR #202, opened 2024-03-13
STOR-1290: don't pass credentials via environment variables in AWS EBS

bertinatto:
CC @openshift/storage


Phaow:
/label qe-approved

Pre merge test with `4.17.0-0.test-2024-07-08-014753-ci-ln-yq665t2-latest`, the `credentials via environment variables` are removed for both standlone and hosted clusters, and the driver works well.

PR #209, opened 2024-04-11
STOR-1726: Add Dockerfile for Samba operator

jsafrane:
Add Dockerfile for Samba operator
@openshift/storage


dobsonj:
CI won't test this dockerfile (yet) so we could override failures if needed.

jsafrane:
/label px-approved
/label docs-approved
/label qe-approved
QE will test the whole operator when it's ready.

PR #215, opened 2024-04-12
STOR-1726: Delete static resources when ClusterCSIDriver is removed

dobsonj:
This PR allows the smb-csi-driver-operator to clean up its static resources when the ClusterCSIDriver is removed.

The approach is the same as https://github.com/openshift/secrets-store-csi-driver-operator/pull/8 (see also: [secrets-store enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/storage/csi-secrets-store.md#uninstalling-csi-driver)). This should allow us to follow a standard approach when moving other OLM operators to csi-operator just by setting `Removable: true` in the `OperatorConfig`.

Operators which are _not_ removable will not define conditional functions, and will behave the same as before with the static resources controller.

Simple manual test:

1) Create ClusterCSIDriver and wait for operator to create static resources:
```
$ oc apply -f - <<EOF
apiVersion: operator.openshift.io/v1
kind: ClusterCSIDriver
metadata:
  name: smb.csi.k8s.io
spec:
  managementState: Managed
EOF
$ oc get pods -n openshift-cluster-csi-drivers
NAME                                             READY   STATUS    RESTARTS   AGE
aws-ebs-csi-driver-controller-7b6cdd676b-745sm   11/11   Running   0          5h55m
aws-ebs-csi-driver-controller-7b6cdd676b-rsnpk   11/11   Running   0          5h55m
aws-ebs-csi-driver-node-5q94r                    3/3     Running   0          5h52m
aws-ebs-csi-driver-node-chs2m                    3/3     Running   0          5h49m
aws-ebs-csi-driver-node-kz24v                    3/3     Running   0          5h55m
aws-ebs-csi-driver-node-q9dj6                    3/3     Running   0          5h55m
aws-ebs-csi-driver-node-qt29w                    3/3     Running   0          5h48m
aws-ebs-csi-driver-node-x2wb6                    3/3     Running   0          5h55m
aws-ebs-csi-driver-operator-5bf867fdd6-27x9f     1/1     Running   0          5h55m
smb-csi-driver-controller-84546d6b8c-cdb75       3/3     Running   0          25s
smb-csi-driver-controller-84546d6b8c-cmgln       3/3     Running   0          25s
smb-csi-driver-node-5dn4m                        3/3     Running   0          23s
smb-csi-driver-node-fg2sc                        3/3     Running   0          23s
smb-csi-driver-node-kb4h6                        3/3     Running   0          23s
smb-csi-driver-node-llnsv                        3/3     Running   0          23s
smb-csi-driver-node-qhfkw                        3/3     Running   0          23s
smb-csi-driver-node-s5c2p                        3/3     Running   0          23s
```

2) Delete ClusterCSIDriver and wait for operator to delete static resources:
```
$ oc delete clustercsidriver smb.csi.k8s.io
clustercsidriver.operator.openshift.io "smb.csi.k8s.io" deleted
$ oc get pods -n openshift-cluster-csi-drivers
NAME                                             READY   STATUS    RESTARTS   AGE
aws-ebs-csi-driver-controller-7b6cdd676b-745sm   11/11   Running   0          5h55m
aws-ebs-csi-driver-controller-7b6cdd676b-rsnpk   11/11   Running   0          5h55m
aws-ebs-csi-driver-node-5q94r                    3/3     Running   0          5h52m
aws-ebs-csi-driver-node-chs2m                    3/3     Running   0          5h49m
aws-ebs-csi-driver-node-kz24v                    3/3     Running   0          5h55m
aws-ebs-csi-driver-node-q9dj6                    3/3     Running   0          5h55m
aws-ebs-csi-driver-node-qt29w                    3/3     Running   0          5h49m
aws-ebs-csi-driver-node-x2wb6                    3/3     Running   0          5h55m
aws-ebs-csi-driver-operator-5bf867fdd6-27x9f     1/1     Running   0          5h56m
```

/cc @openshift/storage @jsafrane @mpatlasov

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/csi/csidrivercontrollerservicecontroller"
 	"github.com/openshift/library-go/pkg/operator/csi/csidrivernodeservicecontroller"
 	"github.com/openshift/library-go/pkg/operator/management"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
@@ -79,6 +80,20 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		controlPlaneControllerInformers = append(controlPlaneControllerInformers, controlPlaneSecretInformer.Informer())
 	}
 
+	// Only removable operators use conditional static resources.
+	// If the operator is not removable, leave these functions nil
+	// to unconditionally sync static resources.
+	var shouldCreateFn resourceapply.ConditionalFunction = nil
+	var shouldDeleteFn resourceapply.ConditionalFunction = nil
+	if opConfig.Removable {
+		shouldCreateFn = func() bool {
+			return getOperatorSyncState(c.OperatorClient) == opv1.Managed
+		}
+		shouldDeleteFn = func() bool {
+			return getOperatorSyncState(c.OperatorClient) == opv1.Removed
+		}
+	}
+
 	controlPlaneCSIControllerSet := csicontrollerset.NewCSIControllerSet(
 		c.OperatorClient,
 		c.EventRecorder,
@@ -92,12 +107,8 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		c.ControlPlaneKubeInformers,
 		a.GetAsset,
 		a.GetControllerStaticAssetNames(),
-		func() bool {
-			return getOperatorSyncState(c.OperatorClient) == opv1.Managed
-		},
-		func() bool {
-			return getOperatorSyncState(c.OperatorClient) == opv1.Removed
-		},
+		shouldCreateFn,
+		shouldDeleteFn,
 	).WithCSIConfigObserverController(
 		csiOperatorControllerConfig.GetControllerName("DriverCSIConfigObserverController"),
 		c.ConfigInformers,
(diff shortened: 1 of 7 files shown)


pkg/operator/starter.go
```diff
@@ -187,3 +202,35 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+// to avoid syncing resources when the actual state is unknown.
+func getOperatorSyncState(operatorClient v1helpers.OperatorClientWithFinalizers) opv1.ManagementState {
+	opSpec, _, _, err := operatorClient.GetOperatorState()
+	if err != nil {
+		klog.Errorf("Failed to get operator state: %v", err)
+		return opv1.Unmanaged
+	}
+	// return the state from the operator if it's not managed
+	if opSpec.ManagementState != opv1.Managed {
+		return opSpec.ManagementState
```
jsafrane, line 223: This will allow users to set `managementState: Removed` for CSI drivers that are not removable and the operator **will** remove.

That's an interesting feature that would allow users to disable e.g. vSphere CSI driver on clusters that don't want it installed. But I think it should be discussed first.

For now and for non-removable CSI drivers, `Removed` should be handled as `Managed` (?) and should throw some sane error. Returning `err` would degrade the cluster. It may be too strong, but it's the easiest thing to do.

dobsonj: Ack, I reworked this in a second commit to maintain the existing behavior.

[NewStaticResourcesController](https://github.com/openshift/library-go/blob/eb2f24c36528fe928b6611b2bf88a734e346fcb6/pkg/operator/staticresourcecontroller/static_resource_controller.go#L117) sets the conditional resource functions to `nil` by default, which is interpreted by [WithConditionalResources](https://github.com/openshift/library-go/blob/eb2f24c36528fe928b6611b2bf88a734e346fcb6/pkg/operator/staticresourcecontroller/static_resource_controller.go#L146-L153) as "always create, never delete". So it's perfectly valid to pass `nil` functions into [WithConditionalStaticResourcesController](https://github.com/openshift/library-go/blob/eb2f24c36528fe928b6611b2bf88a734e346fcb6/pkg/operator/csi/csicontrollerset/csi_controller_set.go#L109-L110) and it will behave exactly as `WithStaticResourcesController` did.

So now removable operators set the conditional functions, and non-removable operators leave them `nil` for the simple static resources controller they always used before.

For removable operators, I changed `getOperatorSyncState` to be consistent with how we manage [static resources in gcp-filestore](https://github.com/openshift/gcp-filestore-csi-driver-operator/blob/2da7613226bf6d36240494484b816c72604c3ad1/pkg/operator/staticresources/staticresources.go#L114). It will not sync the resources if there is an error or if the state is not `Managed`, and it removes them only if `DeletionTimestamp` is set.

PR #225, opened 2024-05-17
OCPBUGS-33776: Adjust port for registrar sidecar in aws-ebs csi driver container

mpatlasov:
Port 10304 collides with aws-efs driver from 4.15. Let's use unused port 10309.


jsafrane:
/lgtm
to fix the issue, however, we need to track our ports better.

PR #232, opened 2024-05-29
STOR-1856: migrate AWS EFS operator to csi-operator

RomanBednar:
This is a preparation step that moves existing operator code to legacy dir under csi-operator. Will be followed by changing CI jobs and ART pipeline to build from this repo instead of the original one.


jsafrane:
https://github.com/openshift/enhancements/blob/master/enhancements/storage/csi-driver-operator-merge.md says that moving to csi-operator is 2 step process and the existing operator should be moved first. Why do you do it in a single step?

RomanBednar:
The bundle dir should not be here yet - removed.

PR #223, opened 2024-04-26
OCPBUGS-32370: Configure scraping of node metrics

jsafrane:
Allow node DaemonSet to expose driver metrics.

I did some refactoring, see individual commits. Either renaming files to have clear `deployment_` / `node_` prefix or making generator code generic for controller / node.

"Add metrics to DaemonSet containers" is the real functionality.

@openshift/storage 

WIP:
* ~pending testing~
* ~update library-go~

PR #233, opened 2024-06-07
Ensure Node Metric Service is created in the openshift-cluster-csi-drivers namespace for both OpenShift and Hypershift

fxierh:


fxierh:
@openshift/storage PTAL, thanks

bryan-cox:
/retitle OCPBUGS-34667: Ensure Node Metric Service is created in the openshift-cluster-csi-drivers namespace for both OpenShift and Hypershift

gnufied:
Okay, this looks alright. Having said that - https://github.com/openshift/csi-operator/pull/223/files PR has other errors and needs to be fixed. Let merge this and we will fix other errors in separate commit.

/approve

PR #240, opened 2024-07-11
STOR-1596: Bump all deps for 4.17

rhrmo:
Updating all dependencies using `go get -u ./...`


go.mod
```diff
@@ -1,25 +1,27 @@
 module github.com/openshift/csi-operator
 
-go 1.21
+go 1.22.0
+
```
gnufied, line 4: I do not think this will work.

go.mod
```diff
@@ -1,25 +1,27 @@
...
-	github.com/openshift/build-machinery-go v0.0.0-20231128094528-1e9b1b0595c8
-	github.com/openshift/client-go v0.0.0-20240405120947-c67c8325cdd8
-	github.com/openshift/library-go v0.0.0-20240517131709-d72e84d1d42e
-	github.com/prometheus/client_golang v1.18.0
-	github.com/spf13/cobra v1.8.0
+	github.com/openshift/api v0.0.0-20240710000542-465787efd0d6
+	github.com/openshift/build-machinery-go v0.0.0-20240419090851-af9c868bcf52
+	github.com/openshift/client-go v0.0.0-20240528061634-b054aa794d87
+	github.com/openshift/library-go v0.0.0-20240709182732-b94141242b0c
+	github.com/prometheus/client_golang v1.16.0
```
jsafrane, line 15: this goes from v1.18.0 to v1.16.0, which is odd. It could be correct, but please try to bump to v1.18.0 and if it fails, then comment why v1.16.0 needs to be used.

go.mod
jsafrane: I think it's OK in go 1.22

go.mod
rhrmo: Thanks for noticing, I re-used the command and did not notice the difference in versions.

PR #251, opened 2024-08-01
STOR-1856: add STS hook for EFS credentials request controller

RomanBednar:
PR https://github.com/openshift/csi-operator/pull/237 for migrating EFS operator to csi-operator repo was missing credentials request hook - as a result EFS operator was not working in STS clusters due to missing secret:

```
MountVolume.SetUp failed for volume "aws-credentials" : secret "aws-efs-cloud-credentials" not found
```

Patch was tested to verify EFS dynamic provisioning on STS cluster and correct handling of role ARN provided via Operator Hub:

1) check provided ARN
```
$ oc -n openshift-cluster-csi-drivers get subscription/aws-efs-csi-driver-operator -o json | jq '.spec.config.env'
[
  {
    "name": "ROLEARN",
    "value": "arn:aws:iam::269733383066:role/rbednar-aws-efs-openshift-cluster-csi-drivers-aws-efs-cloud-cred"
  }
]
```

2) validate ARN propagated to cred. request by the hook
```
$ oc -n openshift-cloud-credential-operator get credentialsrequest/openshift-aws-efs-csi-driver -o json | jq '.spec.providerSpec.stsIAMRoleARN'
"arn:aws:iam::269733383066:role/rbednar-aws-efs-openshift-cluster-csi-drivers-aws-efs-cloud-cred"
```

3) verify secret was created by CCO
```
$ oc -n openshift-cluster-csi-drivers get secret/aws-efs-cloud-credentials
NAME                        TYPE     DATA   AGE
aws-efs-cloud-credentials   Opaque   1      48m
```

4) test dynamic provisioning with EFS
```
$ oc get pvc
NAME    STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
pvc-1   Bound    pvc-32f11702-80db-4c3c-b52a-8ccce73e486a   1Gi        RWO            efs-sc         <unset>                 61s

$ oc get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM           STORAGECLASS   VOLUMEATTRIBUTESCLASS   REASON   AGE
pvc-32f11702-80db-4c3c-b52a-8ccce73e486a   1Gi        RWO            Delete           Bound    default/pvc-1   efs-sc         <unset>                          14s
```

PR #252, opened 2024-08-01
STOR-1993: add init container in EFS CSI controller pod

bertinatto:
CC @openshift/storage

### DIFF
--- a/assets/overlays/aws-efs/generated/standalone/controller.yaml
+++ b/assets/overlays/aws-efs/generated/standalone/controller.yaml
@@ -56,22 +56,10 @@ spec:
         env:
         - name: CSI_ENDPOINT
           value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
-        - name: AWS_ACCESS_KEY_ID
-          valueFrom:
-            secretKeyRef:
-              key: aws_access_key_id
-              name: aws-efs-cloud-credentials
-              optional: true
-        - name: AWS_SECRET_ACCESS_KEY
-          valueFrom:
-            secretKeyRef:
-              key: aws_secret_access_key
-              name: aws-efs-cloud-credentials
-              optional: true
         - name: AWS_SDK_LOAD_CONFIG
           value: "1"
         - name: AWS_CONFIG_FILE
-          value: /var/run/secrets/aws/credentials
+          value: /var/run/aws/auth/credentials
         image: ${DRIVER_IMAGE}
         imagePullPolicy: IfNotPresent
         livenessProbe:
@@ -97,8 +85,8 @@ spec:
           privileged: true
         terminationMessagePolicy: FallbackToLogsOnError
         volumeMounts:
-        - mountPath: /var/run/secrets/aws
-          name: aws-credentials
+        - mountPath: /var/run/aws/auth
+          name: aws-auth
           readOnly: true
         - mountPath: /var/run/secrets/openshift/serviceaccount
           name: bound-sa-token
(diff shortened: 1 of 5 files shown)


assets/overlays/aws-efs/generated/standalone/controller.yaml
```diff
@@ -200,6 +190,24 @@ spec:
...
           name: metrics-serving-cert
       hostNetwork: true
+      initContainers:
+      - command:
+        - sh
+        - -c
+        - |
+          cat <<EOF > /var/run/aws/auth/credentials
+          [default]
+          aws_access_key_id=$(cat /var/run/aws/keys/aws_access_key_id)
```
RomanBednar, line 200: Those files won't exist on STS cluster so it will create a file like:

```
[default]
aws_access_key_id=
aws_secret_access_key=
```

But as long as it does not overwrite STS config it should be fine I guess. Currently the config in STS is at `/var/run/secrets/aws/credentials`, with this patch it would be `/var/run/aws/keys/credentials`. Maybe the script could check the files first?

assets/overlays/aws-efs/patches/controller_add_driver.yaml
```diff
@@ -56,22 +74,10 @@ spec:
...
-              valueFrom:
-                secretKeyRef:
-                  name: aws-efs-cloud-credentials
-                  key: aws_secret_access_key
-                  optional: true
             - name: AWS_SDK_LOAD_CONFIG
               value: '1'
             - name: AWS_CONFIG_FILE
-              value: /var/run/secrets/aws/credentials
+              value: /var/run/aws/auth/credentials
```
RomanBednar, line 80: Isn't this `AWS_CONFIG_FILE` var used by AWS SDK also in case of STS? If so it could break stuff - might be worth testing. However I broke cred. request controller when migrating the code so we might need merging this first before testing STS: https://github.com/openshift/csi-operator/pull/251

bertinatto: Sounds good, let's put an hold to this PR and merge yours first.

bertinatto:
/hold
for https://github.com/openshift/csi-operator/pull/251

assets/overlays/aws-efs/generated/standalone/controller.yaml
bertinatto: Good point, thanks. I fixed this in my other PR for EBS, once that one merges I'll do the same here.

bertinatto:
> /hold for #251

/hold cancel

Phaow:
/label qe-approved
```console
# Checked the presubmit job aws-efs-operator-e2e-extended must-gather the init-aws-credentials-file container works as expected
$ omc logs aws-efs-csi-driver-controller-9d98d9d56-jpt52 -c init-aws-credentials-file
2024-08-13T19:47:37.697578318Z Kubernetes Secret already contains credentials file, copied to the right place: /var/run/aws/auth/credentials
```

config/aws-efs/manifests/stable/image-references
```diff
@@ -27,3 +27,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-kube-rbac-proxy:latest
+  - name: tools
```
Ximinhan, line 30: Hi @bertinatto the key tools needs update, we find build failed due to this reference, for tools it only can be `ose-tools `or `ose-network-tools`

PR #256, opened 2024-08-08
STOR-1992: Enable MSI override for ARO HCP

bryan-cox:


bryan-cox:
We ended up not needing this. CSO is using the same credentials setup for azure CCM config. Hermant, Cesar, and I were able to confirm CSO is using MSI when azure CCM config is setup with MSI. Azure CCM will be setup for MSI in https://github.com/openshift/hypershift/pull/4565

PR #260, opened 2024-08-19
STOR-1819: Distribute OCP specific test manifest with create-efs-volume

jsafrane:
We need the AWS EFS OCP specific manifest available in CI. All other EFS tests manifests are generated by create-efs-volume image, so put this manifest there too.

@openshift/storage


bertinatto (commented):
/lgtm

jsafrane:
/overide ci/prow/e2e-azure-file-nfs-csi
This PR has no effect on azure

PR #272, opened 2024-09-05
OCPBUGS-38922: add tag matching to Azure File storage class

RomanBednar:
This should address two issues:
- prevent matching storage accounts that have private endpoints enabled - this is the original issue in bug report
- prevent matching storage accounts that are not created by our operator - this is a side effect of using the tags, but we intended to address it anyway

The used key/value pair for the tag is unlikely to ever match any other storage account, although still possible if it is tagged the same. Another alternative considered was to hardcode storage account name (`storageAccount` param), but that requires having the storage account created in advance which is non-trivial and would add code to the operator that might not be necessary.

cc @openshift/storage

### DIFF
--- a/assets/overlays/azure-file/base/storageclass.yaml
+++ b/assets/overlays/azure-file/base/storageclass.yaml
@@ -6,7 +6,9 @@ metadata:
 provisioner: file.csi.azure.com
 allowVolumeExpansion: true
 parameters:
+  matchTags: "true"
   skuName: Standard_LRS  # available values: Standard_LRS, Standard_GRS, Standard_ZRS, Standard_RAGRS, Premium_LRS
+  tags: storageClassName=azurefile-csi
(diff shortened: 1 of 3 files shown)


duanwei33:
QE Testing with the open PR: 
- For new installation, the sc azurefile-csi with “tags” and “matchTags” is created, the provisioning and all other features like rezise/clone/snapshot works.
- Install a 4.17 cluster (without the fix yet), create PVCs with sc azurefile-csi which don’t have the “tags” and and “match Tags”, then upgrade to 4.18 with [open PR](https://github.com/openshift/csi-operator/pull/272), after upgrading, the azurefile-csi sc name is still the same but the content changes, the  “tags” and and “match Tags” are added. CSI Driver can still manage the previous PVCs create in 4.17 well like resize/clone/delete, so it should not impact the upgrade.

So the fix looks good from my side.

PR #275, opened 2024-09-13
OSASINFRA-3618: Add Cinder CSI support (part 2)

stephenfin:
This is part 2 of the process of merging the OpenStack Cinder CSI Driver operator into `csi-operator`.

This PR address the fourth step in the [migration guide][migration], namely moving the operator to the new structure in `csi-operator`.

Follow-up PRs will remove the `legacy/openstack-cinder-csi-driver-operator` directory and update both the CI jobs and image builds to remove references to same.

You can compare the assets generated here using the following script, which relies on [difftastic](https://difftastic.wilfred.me.uk/installation.html) (available from the Fedora repos):

```python
#!/usr/bin/env python3

import pathlib
import subprocess
import sys


assets = [
    p
    for p in pathlib.Path('openstack-cinder-csi-driver-operator/assets').glob(
        '*.yaml'
    )
]
assets += [
    p
    for p in pathlib.Path('openstack-cinder-csi-driver-operator/assets/rbac').glob('*.yaml')
]

csi_operator_assets = pathlib.Path(
    'csi-operator/assets/overlays/openstack-cinder/generated/standalone/'
)
for legacy_asset in assets:
    # map assets with different names
    asset_name = {
        'prometheus_rolebinding.yaml': 'prometheus_binding.yaml',
        'lease_leader_election_rolebinding.yaml': 'lease_leader_election_binding.yaml',
    }.get(legacy_asset.name, legacy_asset.name)
    csi_operator_asset = csi_operator_assets / legacy_asset.name
    if not csi_operator_asset.exists():
        print(f'ERROR: {csi_operator_asset} does not exist!')
        sys.exit(1)

    subprocess.run(
        f'difft --ignore-comments {legacy_asset} {csi_operator_asset}',
        shell=True,
    )
```

[migration]: https://github.com/openshift/enhancements/blob/master/enhancements/storage/csi-driver-operator-merge.md#process-of-moving-operators-to-csi-operator-monorepo

**Previous steps:**

- https://github.com/openshift/csi-operator/pull/278
- https://github.com/openshift/csi-operator/pull/281
- https://github.com/openshift/release/pull/56812
- https://github.com/openshift-eng/ocp-build-data/pull/5402
- https://github.com/openshift/openstack-cinder-csi-driver-operator/pull/180


gnufied:
Did you check https://github.com/openshift/enhancements/blob/master/enhancements/storage/csi-driver-operator-merge.md for steps about how to merge operators?

stephenfin:
/retitle OSASINFRA-3618: Add Cinder CSI support (part 2)

stephenfin:
Failures look real. I'm investigating.

pkg/driver/openstack-cinder/openstack_cinder.go
```diff
@@ -0,0 +1,191 @@
...
+
+	generatedAssetBase = "overlays/openstack-cinder/generated"
+)
+
+// GetOpenStackCinderGeneratorConfig returns configuration for generating assets of Cinder CSI driver operator.
+func GetOpenStackCinderGeneratorConfig() *generator.CSIDriverGeneratorConfig {
+	return &generator.CSIDriverGeneratorConfig{
+		AssetPrefix:      "openstack-cinder-csi-driver",
+		AssetShortPrefix: "openstack-cinder",
+		DriverName:       "cinder.csi.openstack.com",
```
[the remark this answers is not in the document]
stephenfin: Oh, yes, we should.

assets/overlays/openstack-cinder/patches/controller_add_driver.yaml
```diff
@@ -0,0 +1,111 @@
...
+  template:
+    metadata:
+      annotations:
+        # FIXME(stephenfin): This was not set previously for openstack-cinder-csi-driver-operator.
+        # but is set in assets/base/controller.yaml. Is settting it to the empty string enough, or
+        # should we remove it from the base and move it into the drivers that need it?
+        cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes: ""
+        # FIXME(stephenfin): This was not set previously for openstack-cinder-csi-driver-operator.
+        # What is the effect/necessity of setting it now?
+        openshift.io/required-scc: privileged
```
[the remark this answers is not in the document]
stephenfin: I can't do that (I did try) because we use host networking for the controller. Not sure why the controller needs host networking but I was copying what was done previously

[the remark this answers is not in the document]
mandre: Yep, not sure. It [was there since the beginning](https://github.com/openshift/openstack-cinder-csi-driver-operator/commit/a6c64477a530b7b132a0a61d75237606b7c4e045#diff-529b166853e98d1126b70e972faa93f1355cb4283af58abf08d9d96509d00cbdR17).

assets/overlays/openstack-cinder/patches/controller_add_driver.yaml
```diff
@@ -0,0 +1,111 @@
...
+    config.openshift.io/inject-proxy: csi-driver
+    config.openshift.io/inject-proxy-cabundle: csi-driver
+spec:
+  template:
+    metadata:
+      annotations:
+        # FIXME(stephenfin): This was not set previously for openstack-cinder-csi-driver-operator.
+        # but is set in assets/base/controller.yaml. Is settting it to the empty string enough, or
+        # should we remove it from the base and move it into the drivers that need it?
+        cluster-autoscaler.kubernetes.io/safe-to-evict-local-volumes: ""
```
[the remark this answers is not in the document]
stephenfin: Done

pkg/driver/common/generator/port_registry.go
```diff
@@ -1,14 +1,14 @@
...
+// different clouds on the same cluster.
 const (
 	AWSEBSLoopbackMetricsPortStart = 8201
 	AWSEBSExposedMetricsPortStart  = 9201
 
 	AWSEFSLoopbackMetricsPortStart = 8211
 	AWSEFSExposedMetricsPortStart  = 9211
 
-	// it should be safe to reuse port for Azure because we do not
-	// expect to deploy azure-disk and EBS driver on same clusters
```
gnufied, line 11: Why did we remove this comment lol?

assets/overlays/openstack-cinder/base/controller_privileged_binding.yaml
```diff
@@ -0,0 +1,14 @@
+# Allow the controller to run privileged pods
```
mandre, line 1: Is that still necessary after https://github.com/openshift/csi-operator/pull/275/commits/c1dc7c651aee697601797a86d0690fc4c0fc38bb ?

stephenfin: Whoops, missed this. Removed.

mandre (commented):
The generated assets are almost identical to what we had in [openstack-cinder-csi-driver-operator](https://github.com/openshift/openstack-cinder-csi-driver-operator), with small differences that appears like bugfixes to me (added timeouts, dropping privileges for the controller, missing kube-rbac-proxy sidecar, ...).

And the openstack CI is happy, so LGTM from a shiftstack standpoint.

/lgtm

rlobillo:
/label qe-approved

Verified through fast-fix procedure. Running verification on the cluster and running openstack-tests (all passing) and cinder-csi tests (all passing but 4 errors due to a known issue https://issues.redhat.com/browse/OCPBUGS-12849).

mandre (commented):
/lgtm

assets/overlays/openstack-cinder/patches/provisioner_add_envvars.yaml
```diff
@@ -0,0 +1,13 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: csi-provisioner
+          env:
+            - name: ENABLE_TOPOLOGY
+              valueFrom:
+                configMapKeyRef:
+                  key: enable_topology
```
gnufied, line 12: What does this do for the csi-provisioner?

pkg/openstack-cinder/util/const.go
```diff
@@ -0,0 +1,9 @@
+package util
+
+const (
+	// Operand and operator run in the same namespace
+	DefaultNamespace         = "openshift-cluster-csi-drivers"
+	OpenShiftConfigNamespace = "openshift-config"
```
gnufied, line 6: So we already have `CSIDriverNamespace` constant in `clients` package in csi-operator, so I would rather reuse that.

pkg/openstack-cinder/config/configsync.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	_, _, err = resourceapply.ApplyConfigMap(ctx, c.kubeClient.CoreV1(), c.eventRecorder, targetConfig)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func translateConfigMap(cloudConfig *v1.ConfigMap, enableTopologyFeature bool) (*v1.ConfigMap, error) {
```
gnufied, line 127: Please a comment describing what this function does.

pkg/driver/openstack-cinder/openstack_cinder.go
```diff
@@ -0,0 +1,205 @@
...
+	hook := csidrivernodeservicecontroller.WithConfigMapHashAnnotationHook(
+		c.ControlPlaneNamespace,
+		cinderConfigName,
+		c.GetControlPlaneConfigMapInformer(c.ControlPlaneNamespace),
+	)
+	informers := []factory.Informer{}
+	return hook, informers
+}
+
+func createConfigMapSyncer(c *clients.Clients) (factory.Controller, error) {
```
gnufied, line 195: add a comment about what we are syncing and why.

pkg/openstack-cinder/config/configsync.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	var sourceConfig *v1.ConfigMap
+
+	// First, we try to retrieve from the Cinder CSI-specific config map
+	sourceConfig, err = c.configMapLister.ConfigMaps(util.OpenShiftConfigNamespace).Get("cinder-csi-config")
+	if err != nil {
+		// Failing that, we attempt to retrieve from the cloud provider-specific config map
+		if errors.IsNotFound(err) {
+			sourceConfig, err = c.configMapLister.ConfigMaps(util.OpenShiftConfigNamespace).Get(infra.Spec.CloudConfig.Name)
```
gnufied, line 101: nit: some of the nesting here can be simplified by using following pattern:

```
		if !errors.IsNotFound(err) {
			return err
		}
```

pkg/openstack-cinder/config/configsync.go
```diff
@@ -0,0 +1,220 @@
...
+			if err != nil {
+				if errors.IsNotFound(err) {
+					// TODO: report error after some while?
+					klog.V(2).Infof("Waiting for config map %s from %s", infra.Spec.CloudConfig.Name, util.OpenShiftConfigNamespace)
+					return nil
```
gnufied, line 106: So, if we couldn't find configmap and assuming nothing else happend, we won't know we have an error until next 20 mins? Is that what we want?

assets/overlays/openstack-cinder/generated/standalone/controller.yaml
```diff
@@ -0,0 +1,366 @@
...
+          items:
+          - key: cloud.conf
+            path: cloud.conf
+          name: cloud-conf
+        name: config-cinderplugin
+      - configMap:
+          items:
+          - key: ca-bundle.pem
+            path: ca-bundle.pem
+          name: cloud-provider-config
```
gnufied, line 364: we are calling ca-bundle volume `cloud-provider-config` ?

assets/overlays/openstack-cinder/patches/provisioner_add_envvars.yaml
stephenfin: This is the mechanism we've used to allow us to enable/disabled topology support, which is documented [here](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/features.md#topology). By way of more context, the feature can only be enabled or disabled via a CLI flag (as opposed to something in the configuration file). This meant we needed a way to get the configuration from a user-defined config map (which is how we're storing the rest of our configuration) to that CLI. Using an env var was the solution we arrived in https://github.com/openshift/openstack-cinder-csi-driver-operator/pull/127. There may be better ways to do this and if so, I'm open to them, though we should probably do that separately, in a follow-up PR, in order to keep the asset diff as small as possible here.

pkg/openstack-cinder/util/const.go
stephenfin: Ack, done.

pkg/openstack-cinder/config/configsync.go
stephenfin: Done, although as I highlighted in the commit messages, this code is a direct copy-paste from `openstack-cinder-csi-driver-operator` and is intentionally unchanged insofar as that's possible, to avoid making the review more complicated than necessary.

pkg/driver/openstack-cinder/openstack_cinder.go
stephenfin: Done

pkg/openstack-cinder/config/configsync.go
stephenfin: No, it's not what we want and it's something we should improve, but I didn't figure out a better way to do this at the time (https://github.com/openshift/openstack-cinder-csi-driver-operator/pull/78). I'm open to suggestions but while this is not ideal, it is working so I'd like to keep the fix to a separate PR just to avoid swelling this any further.

assets/overlays/openstack-cinder/generated/standalone/controller.yaml
stephenfin: No, this is in the context of a volume mapping, so `name` here refers to the name of the config map that we wish to extract data from and use to populate a volume. The `cloud-provider-config` config map (CM) may contain 2 keys: `config`, which is required and contains the config for the CSI driver, and `ca-bundle.pem`, which is optional and (if defined) contains a CA bundle for the cloud. Here we're extracting the value of the `ca-bundle.pem` CM key to a volume called `cacert`, which gets mounted to `/etc/kubernetes/static-pod-resources/configmaps/cloud-config` (per the `volumeMounts` argument for the `csi-driver` container).

stephenfin:
@gnufied I have an ask: assuming there is nothing too egregious here, would you be okay with me addressing the comments here in a follow-up PR?

assets/overlays/openstack-cinder/patches/provisioner_add_envvars.yaml
gnufied: Topology feature is enabled by default in external-provisioner. AFAIK - this environment variable in csi-provisioner does nothing. cc @jsafrane

stephenfin: We discussed this elsewhere, but the environment flag doesn't do anything by itself. Rather, it's used to provide the value to the `--feature-gates=Topology=XXX` value, as seen [here](https://github.com/openshift/csi-operator/pull/275/files#diff-fb444a9cea9026f12ee5dced1c8bc96685085720fcb3b2e307f20a09a2a7d1baR140-R147).

There is an issue here in that this feature gate is going away soon so we will need a new strategy. I'm going to work on that this cycle, but I don't think we should hold up merging this for that since this is effectively maintaining the status quo. https://github.com/openshift/csi-operator/blob/61c9fee3acd6fe92e681ac1b63dfc6dcb1d4df54/legacy/openstack-cinder-csi-driver-operator/assets/controller.yaml#L100-L115 Also, we *have* to figure out how to do this properly this cycle since we're going to lose this feature otherwise.

PR #279, opened 2024-09-17
NO-JIRA: Move doc on migrating CSI Driver in-tree

stephenfin:
This is far more discoverable and it allows us to iterate of the procedure faster. The enhancement will be updated to reference this new doc.


gnufied:
I am personally fine with this. @openshift/team-storage-committers what do you think?

gnufied:
@stephenfin can you check failing cinder-csi tests. If they are going to block further merges into csi-operator repo, it may be worth making them optional.

gnufied:
@stephenfin I think there was a way to test this stuff. Can you please file a bug for this? I wonder if those path sepatrors needs to be escaped. This is going to be an issue in merging stuff into `csi-operator` if this remains broken.

PR #283, opened 2024-09-23
OCPBUGS-42325: Add ARO HCP Managed Identity Sidecar Containers

bryan-cox:
This commit adds the adapter-init and adapter-server sidecar containers to the azure-disk-csi-controller and azure-file-csi-controller. These containers are needed for ARO HCP to authenticate with managed identities with Azure API.


gnufied:
How are we planning to handle the fact that, this will require pod to be privileged now? Do you think it is worth updating an existing OCP enhancement for these changes?

bryan-cox:
/hold

Microsoft is requesting to not use the sidecar containers.

PR #285, opened 2024-09-26
NO-JIRA: Add SidecarConfig.WithPatches

stephenfin:
To allow us to modify aspects of the default asset beyond 'args' without having to define a wholly new asset. We also add docs while we're here, mainly to emphasise that this shouldn't actually be used unless necessary.

### DIFF
--- a/pkg/generator/types.go
+++ b/pkg/generator/types.go
@@ -173,6 +173,12 @@ func (cfg SidecarConfig) WithAdditionalAssets(assets ...string) SidecarConfig {
 	return newCfg
 }
 
+func (cfg SidecarConfig) WithPatches(flavours sets.Set[ClusterFlavour], namePatchPairs ...string) SidecarConfig {
+	newCfg := cfg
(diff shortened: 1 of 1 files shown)


jsafrane:
/retest-required
@stephenfin do you have an example of a CSI sidecar patch? Can you mention it in the commit message?

stephenfin:
Yes, I'm using this in https://github.com/openshift/csi-operator/pull/275. You can see it used [here](https://github.com/openshift/csi-operator/pull/275/commits/c7c58c18207f0fc3ad143f53a2d35d5f57a3cc60#diff-9e4d5abc381c99fd5db77cd5b530830b31ceba8d952f57da3f095fa3bd7fcac8R62-R64).

jsafrane:
/label px-approved
/label docs-approved
/label qe-approved
This is not an user-facing change + QE will be done as part of https://github.com/openshift/csi-operator/pull/275 (or the PR that switches Cinder operator image to the csi-operator based one.

PR #286, opened 2024-09-26
NO-JIRA: Rename efscreate package

stephenfin:
Align with the name of the binary as is done for everything else.

Signed-off-by: Stephen Finucane <stephenfin@redhat.com>


gnufied:
I think if we do this, we will have to update how we run e2e tests for efs driver too in `openshift/release` repo.

stephenfin:
I'll revisit this later.

PR #287, opened 2024-09-26
NO-JIRA: Remove generated assets before generation

stephenfin:
This ensures we are not left with stale resources...like the four we are deleting here!


jsafrane:
/label px-approved
/label docs-approved
/label qe-approved
removing an unused file.

PR #290, opened 2024-10-07
OCPBUGS-42325: Set separate secrets for file and disk config on HyperShift

bryan-cox:
Set separate secrets for file and disk authentication config for Azure. Previously, these two components were reusing the same configuration as cloud provider.

### DIFF
--- a/assets/overlays/azure-disk/generated/hypershift/controller.yaml
+++ b/assets/overlays/azure-disk/generated/hypershift/controller.yaml
@@ -409,4 +409,4 @@ spec:
           secretName: service-network-admin-kubeconfig
       - name: cloud-config
         secret:
-          secretName: azure-cloud-config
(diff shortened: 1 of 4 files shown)


jsafrane:
It's a good idea to separate these secrets, is there a corresponding hypershift PR that _creates_ those secrets?

bryan-cox:
> It's a good idea to separate these secrets, is there a corresponding hypershift PR that _creates_ those secrets?

Hey @jan--f - there isn't a PR at the moment but there will be in the near future. HyperShift would be responsible for creating/reconciling that within the HCP namespace.

bryan-cox:
I opened https://github.com/openshift/hypershift/pull/4861 for now to correlate to this PR.

Phaow:
Will do some intergration test with hypershift QE guys later when the configuration of cluster could be created.

PR #297, opened 2024-10-11
CFE-1131: AWS Tags DAY2 Update

anirudhAgniRedhat:
This PR introduces a custom EBSVolumeTagController that monitors the OpenShift Infrastructure resource for changes in AWS ResourceTags. When tags are updated, the controller automatically fetches all AWS EBS-backed PersistentVolumes (PVs) in the cluster, retrieves their volume IDs, and updates the associated EBS tags in AWS.

Key Changes:

Monitors Infrastructure resource for AWS ResourceTags updates.
Directly fetches all PVs using the AWS EBS CSI driver (ebs.csi.aws.com).
Updates AWS EBS tags by merging new and existing tags using the AWS SDK.


anirudhAgniRedhat:
/hold 
Please don't review now currently WIP

anirudhAgniRedhat:
/unhold 
open for reviews

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,338 @@
...
+		return err
+	}
+
+	// Merge the existing tags with new resource tags
+	mergedTags := mergeTags(existingTagsOutput.Tags, resourceTags)
+
+	klog.Infof("Updating EBS tags for volume ID %s with tags: %v", volumeID, mergedTags)
+
+	// Create or update the tags
+	_, err = ec2Client.CreateTags(&ec2.CreateTagsInput{
```
TrilokGeer, line 284: What is the behaviour when no.of tags > 50?

anirudhAgniRedhat:
/cc @jsafrane 
PTAL!!

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: Hey @TrilokGeer I tested the scenario the AWS client gives a bad request error which states `Error updating tags for volume vol-08d7a36e4b077873f: TagLimitExceeded: More than 50 tags specified
	status code: 400, request id: 364561fd-6319-40e2-b3d1-ff110954e73c` 
	
We are not reconciling again on this request. What else you think we should do other than this??

jsafrane (commented):
Please update cluster-storage-operator to add token-minter sidecar.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
TrilokGeer: An error event or message must be set on the operator status to indicate the error.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,338 @@
...
+
+// handleDelete is called when an Infrastructure resource is deleted
+func (c *EBSVolumeTagController) handleDelete(obj interface{}) {
+	infra := obj.(*configv1.Infrastructure)
+	klog.Infof("Infrastructure resource deleted: %s", infra.Name)
+}
+
+// processInfrastructure processes the Infrastructure resource and updates EBS tags
+func (c *EBSVolumeTagController) processInfrastructure(infra *configv1.Infrastructure) {
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil &&
```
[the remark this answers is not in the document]
anirudhAgniRedhat: https://github.com/openshift/csi-operator/pull/297/files#diff-dff3b1f75885ac7047cf6b6de01bc2b47b96876def7e60f08b41ce25232c8148R93

Added above!!

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,338 @@
...
+				Values: []*string{aws.String(volumeID)},
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	// Merge the existing tags with new resource tags
+	mergedTags := mergeTags(existingTagsOutput.Tags, resourceTags)
```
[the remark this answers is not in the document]
anirudhAgniRedhat: Yes correct!! I have removed the describe Function, CreateTags is merging on its own.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,446 @@
...
+		name:          name,
+		commonClient:  commonClient,
+		eventRecorder: eventRecorder,
+		failedQueue:   workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+	}
+	// Define the informer for watching Infrastructure resource changes
+	infraInformer := c.commonClient.ConfigInformers.Config().V1().Infrastructures().Informer()
+
+	// Add event handler to process updates only when ResourceTags change
+	infraInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
gnufied, line 58: We don't have do this. If we are using `factory.WithInformers`, we will reconcile this within `Sync` function.

anirudhAgniRedhat: Hey @gnufied I was bit confused with this!! basically I don't want to run reconciliation on every change on InfraStructure resource. I need to run reconciliation only if there is a any change in `infra.Status.PlatformStatus.AWS.ResourceTags`.

Can you suggest a better way to do this so that we can remove the unnecessarily computes.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,483 @@
...
+	}
+	var infraRegion = ""
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil {
+		infraRegion = infra.Status.PlatformStatus.AWS.Region
+	}
+	ec2Client, err := c.getEC2Client(ctx, infraRegion)
+	if err != nil {
+		return err
+	}
+	err = c.processInfrastructure(infra, ec2Client)
```
gnufied, line 117: So, does this controller needs to be an opt-in or a default controller? I don't assume every OCP customer wants this feature and if tagging were to fail after OCP upgrade, their clusters will be degraded and we will have support nightmare.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,483 @@
...
+func volumesIDsToResourceIDs(volumeIDs []string) []*string {
+	var resourceIDs []*string
+	for _, volumeID := range volumeIDs {
+		resourceIDs = append(resourceIDs, aws.String(volumeID))
+	}
+	return resourceIDs
+}
+
+// startFailedQueueWorker runs a worker that processes failed batches independently
+func (c *EBSVolumeTagsController) startFailedQueueWorker(ctx context.Context) {
```
gnufied, line 315: I am not sure if we need this function at all.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
gnufied: cc @jsafrane

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: > How will this work when controller is restarted? You are also doing `WithInformers` below and hence any change in `infra` object will still trigger `Sync`.

What I am thinking is on every restart I would like to run a sync function and update the volumes if there is a change and further On each change in resource-tags we would again run the sync.


> So, what we have currently is worst of both the worlds. If I were to design this, I will probably make a hash of sorted tags and annotate PV with tag hash. If tag hash annotation in PV and computed tag hash don't change, then I will not update the PV or else I will.

This looks easy way to manage this!! I will then add a new field in controller struct which will have the updated hash for the sorted map of tags! now in each reconciliation I will update tags only if the hash is different from the other one? Does this sounds better to you??

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: So the reason I brought this change is I would like to retry the the batches which have failed to update tags!! Here I would like to update tags in a serial order(one By one) [Discussuion link](https://redhat-internal.slack.com/archives/C057L1GKYSK/p1729528491363079) for the volumes. I cannot use the similar sync function here or else you could suggest a better way for this!!

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,511 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const (
+	awsSecretNamespace     = "openshift-cluster-csi-drivers"
+	awsSecretName          = "ebs-cloud-credentials"
+	infrastructureResource = "cluster"
+	driverName             = "ebs.csi.aws.com"
+	tagHashAnnotationKey   = "ebs.csi.aws.com/volume-tags-hash"
```
gnufied, line 38: we should keep this in openshift namespace. So something like - `ebs.openshift.io/volume-tags-hash`

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
gnufied: If we are going to per-pv hash, then we are not going to need this right?

anirudhAgniRedhat: The points I want to make sure here are 

1. I definitely need to batch volumes in order to not to hit throttling condition.
2. Now on failure we would need to retry!!
3. Since we are using batch APIs, so AWS SDK's APIs will give us error if any one of the VolumeID in the batch hits the any error(May be validation, Auth, Permission etc), All the VolumeIDs in that batch will not be able to update the tags! In this condition I would like to add a worker queue that will handle the the serial update of PV's tags and will retry to update the tags in exponential back-off time-period.

Since We need to figure out the trade-off between the either in first place we should update all PVs using AWS API's in serial order and retry using similar sync function or should we use Batch Volumes to be called in the sync Functions and later retry should be processed with the queue function serially until the queue is empty!!

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: Added Thanks For Suggestion!!

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,528 @@
...
+	eventRecorder events.Recorder) factory.Controller {
+
+	c := &EBSVolumeTagsController{
+		name:          name,
+		commonClient:  commonClient,
+		eventRecorder: eventRecorder,
+		failedQueue:   workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+	}
+
+	go c.startFailedQueueWorker(ctx)
```
gnufied, line 61: Problem with starting a worker here is, we are not waiting for caches to be synced before starting stuff. We shouldn't be starting workers at this stage.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
gnufied: >I definitely need to batch volumes in order to not to hit throttling condition.

That is fine. You already batching PVs in `fetchPVsAndUpdateTags`.

>Now on failure we would need to retry!!

We are unnecessarily complicating this. The controller is going to `resync` every 20 minutes anyways, so it will try to tag all the PVs which aren't tagged. So do we even need to keep separate queue for failed PVs? I am also afraid that, your failed worker queue is going to race with normal controller resync.

anirudhAgniRedhat: Hey @gnufied I completely agree that this retrying the failure is unnecessarily making the changes complicated.

@jsafrane I had a chat with @TrilokGeer regarding this, IMHO we should drop the idea to add degraded condition on tags update failure. As the cluster should not be degraded for failure to apply Tags. as the sync will anyhow retry to tag the volumes withing resync period. 
Also anyways we are emitting the warning events from the this controller if the tags update is failed. so user will know that the tags update has been failed due the the certain reason.

/cc @TrilokGeer 
Can you also put your views on this!!

gnufied: BTW are we planning to backport this PR to older releases?

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
jsafrane: It is presented as enabled by default + no opt out for all HyperShift clusters in the [enhancement](https://github.com/openshift/enhancements/pull/1700).

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,528 @@
...
+			klog.Errorf("Error processing PVs for infrastructure update: %v", err)
+			return err
+		}
+	}
+	return nil
+}
+
+// fetchPVsAndUpdateTags retrieves all PVs and updates the AWS EBS tags in batches of 100
+func (c *EBSVolumeTagsController) fetchPVsAndUpdateTags(resourceTags []configv1.AWSResourceTag, ec2Client *ec2.EC2) error {
+	pvs, err := c.commonClient.KubeClient.CoreV1().PersistentVolumes().List(context.TODO(), metav1.ListOptions{})
```
gnufied, line 313: We should be using informers here.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: @gnufied I guess we would need to backport this to 4.17. [Slack Thread](https://redhat-internal.slack.com/archives/C057L1GKYSK/p1727275576645219?thread_ts=1727193184.791609&cid=C057L1GKYSK).

gnufied: So I noticed that you removed failed worker logic. The thing is - I was talking to @jsafrane offline and he has me convinced that, we do need some kind of additional logic so as we can try retagging of "failed" PVs one-by-one, rather than in a batch. This will ensure that, one bad apple in a batch doesn't prevent tagging of rest of the PVs.

But - we need to be careful when doing this. 
- We should make sure that, PVs which will be tagged via failed worker, doesn't get processed via regular controller resync (so no race). 
- I would move the entire failed worker code in a separate file.

anirudhAgniRedhat: Ack; 
I believe we can remove the  `ResyncEvery` parameter from the controller builder as this is a overkill for us why do we want to run the sync function in every 20 mins if nothing has changed in the resource Tags.

Alternatively, If you really think resync is important here then, I think that we can add the handle the race condition by using another annotation in PVs for Tagging status and filtering based on tagshash and status in the annotation. But this will also cost us some volume Update calls and further need to handle cases where we are not able to update the status within batchs.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: I have removed the worker from here and we are initializing the worker in postStartHook after the caches are synced and controller is started.

anirudhAgniRedhat:
> Please update cluster-storage-operator to add token-minter sidecar.

Added PR to add token minter sidecar PTAL!
 https://github.com/openshift/cluster-storage-operator/pull/528

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+
+	if _, err := f.Write(data); err != nil {
+		defer os.Remove(f.Name())
+		return "", fmt.Errorf("failed to write credentials to %s: %v", f.Name(), err)
+	}
+	return f.Name(), nil
+}
+
+// getInfrastructure retrieves the Infrastructure resource in OpenShift
+func (c *EBSVolumeTagsController) getInfrastructure(ctx context.Context) (*configv1.Infrastructure, error) {
```
gnufied, line 192: why are we reading an object from informer with exponential backoff? See how we do this in other operators -  https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/master/pkg/operator/vspherecontroller/vspherecontroller.go#L167

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+
+	infra, err := c.getInfrastructure(ctx)
+	if err != nil {
+		return err
+	}
+	var infraRegion = ""
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil {
+		infraRegion = infra.Status.PlatformStatus.AWS.Region
+	}
+	ec2Client, err := c.getEC2Client(ctx, infraRegion)
```
gnufied, line 94: IMO, we shouldn't create `ec2Client` here until we know we are going to need to tag PV objects. Since we are planning to backport this feature, for most infra objects, there may not even be *any* tags that needded to be synced on AWS. So, this controller will be a NO-OP for many clusters.

This means at very minimum, I would check if `infra` object has non-nil `infra.Status.PlatformStatus.AWS.ResourceTags`, before going ahead with ec2 client creation.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+		Tags:      mergedTags,
+	})
+
+	if err != nil {
+		return fmt.Errorf("error creating tags for volume %v: %v", pvBatch, err)
+	}
+	return nil
+}
+
+func (c *EBSVolumeTagsController) listPersistentVolumesWithRetry(ctx context.Context) ([]*v1.PersistentVolume, error) {
```
gnufied, line 318: same comment as before. There is no need to list PV objects from a informer with an exponential backoff. Informer has cached objects, it isn't supposed to fail.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+	// Emit a warning event for the failure
+	c.eventRecorder.Warning("EBSVolumeTagsUpdateFailed", fmt.Sprintf("Failed to update tags for batch %v: %v", pvNames, errorMessage))
+}
+
+// newAndUpdatedTags adds and update existing AWS tags with new resource tags from OpenShift infrastructure
+func newAndUpdatedTags(resourceTags []configv1.AWSResourceTag) []*ec2.Tag {
+	// Convert map back to slice of ec2.Tag
+	var tags []*ec2.Tag
+	for _, tag := range resourceTags {
```
gnufied, line 373: I am confused. The function definition says - it is supposed to "merge" tags, but all I see is, this function returns `configv1.AWSResourceTag` as `ec2.Tag` objects. I do not see any "merge" happening.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+		}
+		return true, nil
+	})
+	return err
+}
+
+func (c *EBSVolumeTagsController) handleTagUpdateFailure(batch []*v1.PersistentVolume, error error) {
+	errorMessage := fmt.Sprintf("Error updating tags for volume %v: %v", batch, error)
+	for _, pv := range batch {
+		klog.Errorf("Error updating volume %v tags: %v", pv.Name, errorMessage)
```
gnufied, line 358: so, I do not think dumping full PV objects in error log is a good idea here. This will make logs very noisy. Also, didn't we add full PV object already in `error` object?

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+	mergedTags := newAndUpdatedTags(resourceTags)
+
+	// Create or update the tags
+	_, err := ec2Client.CreateTags(&ec2.CreateTagsInput{
+		Resources: pvsToResourceIDs(pvBatch),
+		Tags:      mergedTags,
+	})
+
+	if err != nil {
+		return fmt.Errorf("error creating tags for volume %v: %v", pvBatch, err)
```
gnufied, line 313: Aren't we appending full PV objects again and again to same error message?

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
gnufied: Please do not use `error` as a variable name, since `error` is also an interface. This confuses tools.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+
+// setPVTagHash stores the hash in the PV annotations.
+func setPVTagHash(pv *v1.PersistentVolume, hash string) {
+	// Ensure the PV has an annotations map
+	if pv.Annotations == nil {
+		pv.Annotations = make(map[string]string)
+	}
+
+	// Set or update the tag hash annotation
+	pv.Annotations[tagHashAnnotationKey] = hash
```
gnufied, line 411: We should get `DeepCopy` before modifying object we retrieved from informer. We do not want to modify object we retrieved from informers directly.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+
+		// Update PV annotations after successfully updating the tags in AWS
+		for _, volume := range batch {
+			// Set the new tag hash annotation in the PV object
+			setPVTagHash(volume, newTagsHash)
+
+			// Update the PV with the new annotations
+			err = c.updateVolumeWithRetry(ctx, volume)
+			if err != nil {
+				klog.Errorf("Error updating PV annotations for volume %s: %v", volume.Name, err)
```
gnufied, line 290: If we are already going to handle these PVs via `failedWorkerQueue`, why does the code in `updateVOlumeWithRetry` needs to have its own local exponential backoff?

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+	newTagsHash := computeTagsHash(resourceTags)
+	pvsToBeUpdated := filterUpdatableVolumes(pvs, newTagsHash)
+
+	// If there are no volumes to update, return early
+	if len(pvsToBeUpdated) == 0 {
+		klog.Infof("No volume tags to update as hashes are unchanged")
+		return nil
+	}
+
+	// Process the volumes in batches
```
gnufied, line 267: I mentioned in my original comment that, the main controller should not try and update PVs which will be processed by failed worker queue. I do not see that scenairo handled here.

pkg/driver/aws-ebs/aws_ebs_tags_retry_worker.go
```diff
@@ -0,0 +1,101 @@
...
+// startFailedQueueWorker runs a worker that processes failed volumes independently
+func (c *EBSVolumeTagsController) startFailedQueueWorker(ctx context.Context, syncContext factory.SyncContext) error {
+	for {
+		select {
+		case <-ctx.Done():
+			klog.Infof("Context canceled, stopping failed queue worker for EBS Volume Tags")
+			return errors.New("context canceled, stopping failed queue worker for EBS Volume Tags") // Stop the goroutine when the context is canceled
+		default:
+			// Get the next failed volume from the queue
+			item, quit := c.failedQueue.Get()
```
gnufied, line 22: This code is very hard to read, can you split this into multiple smaller functions please.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,441 @@
...
+	backoff := wait.Backoff{
+		Duration: operationDelay,
+		Factor:   operationBackoffFactor,
+		Steps:    operationRetryCount,
+	}
+	var awsCreds *v1.Secret
+	err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
+		var apiError error
+		awsCreds, apiError = c.commonClient.KubeClient.CoreV1().Secrets(awsEBSSecretNamespace).Get(ctx, awsEBSSecretName, metav1.GetOptions{})
+		if apiError != nil {
```
gnufied, line 224: Can we use a informer/lister here as well?

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: Updated the definition, Thanks for pointing out

pkg/driver/aws-ebs/aws_ebs_tags_retry_worker.go
```diff
@@ -0,0 +1,121 @@
...
+	pvName, ok := obj.(string)
+	if !ok {
+		klog.Errorf("Invalid volume name type in failed queue, skipping")
+		c.failedQueue.Forget(obj)
+		return
+	}
+
+	klog.Infof("Retrying failed volume: %v", pvName)
+
+	infra, err := c.getInfrastructure()
```
TrilokGeer, line 45: Race condition: As I understand from @anirudhAgniRedhat , the worker is a separate thread running concurrently. In that case, the call here does not gurantee that the infrastructure being read is latest. There is probability of older version infrastructure tags and updating to the pv. As this operation is not iterated until next update, it will lead to a situation where some of the volumes carry older versions of tags.

pkg/driver/aws-ebs/aws_ebs_tags_retry_worker.go
```diff
@@ -0,0 +1,121 @@
...
+			c.failedQueue.Forget(pvName)
+			return
+		}
+		klog.Errorf("Failed to get persistent volume %v: %v", pvName, err)
+		c.failedQueue.AddRateLimited(pvName)
+		return
+	}
+
+	if c.needsTagUpdate(infra, pv) {
+		c.updateTags(ctx, pv, infra.Status.PlatformStatus.AWS.Region, infra.Status.PlatformStatus.AWS.ResourceTags)
```
TrilokGeer, line 70: Based on the earlier assumption about worker queue being a concurrent thread, update tags becomes a critical section between sync and worker. To avoid race conditions, it is best to consider a single thread to have update call functionality.

pkg/driver/aws-ebs/aws_ebs_tags_retry_worker.go
```diff
@@ -0,0 +1,121 @@
...
+
+	err = c.updateVolume(ctx, updatedVolume)
+	if err != nil {
+		klog.Errorf("Error updating PV annotations for volume %s: %v", pv.Name, err)
+		c.failedQueue.AddRateLimited(pv.Name)
+		return
+	}
+
+	klog.Infof("Successfully updated PV annotations for volume %s", pv.Name)
+	c.failedQueue.Forget(pv.Name)
```
TrilokGeer, line 120: Maybe, there should be a consideration to provide a status to user about an in-progress activity. With no updates, user cannot know if the tags have been applied successfully, failed or in-progress.

TrilokGeer: Maybe a separate set of events might help?

TrilokGeer:
AFAIU, it is good to consider 
1. Sync loop blocking time on tag updates to optimize the success cases.
2. For error cases, tag data consistency is required between sync and worker updates. Reading infrastructure object in both threads and independent updates leads to race conditions.
3. On retry for failed PVs, there is no feedback or consideration to skip the failing PV in next batch command. This might result in iterating the PVs again.
4. A way to inform about the status of the present tag update action.
@jsafrane @gnufied @anirudhAgniRedhat

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+func (c *EBSVolumeTagsController) filterUpdatableVolumes(volumes []*v1.PersistentVolume, newTagsHash string) []*v1.PersistentVolume {
+	var updatablePVs []*v1.PersistentVolume
+
+	for _, volume := range volumes {
+		// Check if the volume is a CSI volume with the correct driver
+		if volume.Spec.CSI != nil && volume.Spec.CSI.Driver == driverName &&
+			// Ensure the volume is not already in the failed queue
+			!c.isVolumeInFailedQueue(volume.Name) &&
```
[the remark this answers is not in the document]
anirudhAgniRedhat: Yes @TrilokGeer, Here we cannot update the tags immediately if the volume is in a failed queue.
But it would be even worse if we try to refresh the queue by removing all volume from the queue and then retry to update the tags in batches. as we are calling the sync func here when there is any update on Infrastructure resource.

pkg/driver/aws-ebs/aws_ebs_tags_retry_worker.go
anirudhAgniRedhat: Correct maybe we would need to resync in every x mins, if the volume is out of sync due to reading older data here.

WDYT @gnufied @jsafrane ?

gnufied: I do think we need to resync every x minutes, since we are not even watching for PV objects. Any sufficient x minute resync will fix errors like these.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,422 @@
...
+
+	awsErrorVolumeNotFound = "InvalidVolume.NotFound"
+)
+
+type EBSVolumeTagsController struct {
+	name          string
+	commonClient  *clients.Clients
+	eventRecorder events.Recorder
+	failedQueue   workqueue.TypedRateLimitingInterface[string]
+	rateLimiter   *rate.Limiter
```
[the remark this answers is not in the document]
anirudhAgniRedhat: https://github.com/openshift/csi-operator/pull/297/files#diff-dff3b1f75885ac7047cf6b6de01bc2b47b96876def7e60f08b41ce25232c8148R282 

We are using the rate limitter  here to limit the API calls based on bucket token method used in AWS.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,422 @@
...
+		name:          name,
+		commonClient:  commonClient,
+		eventRecorder: eventRecorder,
+		failedQueue:   workqueue.NewTypedRateLimitingQueue[string](workqueue.NewTypedItemExponentialFailureRateLimiter[string](10*time.Second, 100*time.Hour)),
+		rateLimiter:   rateLimiter,
+	}
+	return factory.New().WithSync(
+		c.Sync,
+	).WithInformers(
+		c.commonClient.ConfigInformers.Config().V1().Infrastructures().Informer(),
```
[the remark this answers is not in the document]
anirudhAgniRedhat: https://github.com/openshift/csi-operator/blob/main/pkg/driver/aws-ebs/aws_ebs.go#L372

For the newer volume resources the above deployment hook is able to add the tags on the newer resources.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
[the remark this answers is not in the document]
anirudhAgniRedhat: I just wanted to ensure we don't end up hitting more APIs for now. Its better to keep a limiter that would help us in this senerio.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,422 @@
...
+			updatablePVs = append(updatablePVs, volume)
+		}
+	}
+	return updatablePVs
+}
+
+// isVolumeInFailedQueue checks if a volume name is currently in the failed queue
+func (c *EBSVolumeTagsController) isVolumeInFailedQueue(volumeName string) bool {
+	// Check if the volume name is in the failed queue
+	return c.failedQueue.NumRequeues(volumeName) > 0
```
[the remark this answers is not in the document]
anirudhAgniRedhat: Sorry I didn't understood your concern here. Can you please help me why do you think this would not look for volumeName here?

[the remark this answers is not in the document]
anirudhAgniRedhat: https://github.com/openshift/csi-operator/blob/main/vendor/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L36

Here the NumRequeues func  should tells us that how many times the object is failed. we assume if the number of retries are more that 0 and object is still in the queue. this means we can leverage the retry queue worker to update the tags for the particular resource.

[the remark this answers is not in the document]
gnufied: I have tested this code in a standalone snippet and what you are doing will only work if tagging of a volume fails in failed worker queue. If you add an item for the first time in failed worker queue, then `NumRequest` will return 0.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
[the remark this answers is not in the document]
gnufied: ack. Should be okay I guess.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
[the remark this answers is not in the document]
anirudhAgniRedhat: Correct @gnufied This can happen on 2 concurrent calls within the first fail reconciliation for tag update.

Do you think we can leverage a new field for identifying the volumes which are in retry queue ?
```
type EBSVolumeTagsController struct {
	failedQueue workqueue.RateLimitingInterface
	failedSet   map[string]struct{} // A set to track added volume names
	mutex       sync.Mutex          // To prevent concurrent map access
}

// isVolumeInFailedQueue checks if a volume name is currently in the failed queue
func (c *EBSVolumeTagsController) isVolumeInFailedQueue(volumeName string) bool {
	c.mutex.Lock()
	defer c.mutex.Unlock()

	// Check if the volume name is in the set
	_, exists := c.failedSet[volumeName]
	return exists
}

// addToFailedQueue adds a volume name to the failed queue and tracks it in the set
func (c *EBSVolumeTagsController) addToFailedQueue(volumeName string) {
	c.mutex.Lock()
	defer c.mutex.Unlock()

	// Add volume name to the queue and set
	c.failedQueue.Add(volumeName)
	c.failedSet[volumeName] = struct{}{}
}

// removeFromFailedQueue removes a volume name from the queue and the set
func (c *EBSVolumeTagsController) removeFromFailedQueue(volumeName string) {
	c.mutex.Lock()
	defer c.mutex.Unlock()

	// Remove volume name from the queue and set
	c.failedQueue.Forget(volumeName)
	delete(c.failedSet, volumeName)
}
```

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,424 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating AWS session using credentials file: %v", err)
+	}
+
+	return ec2.New(awsSession), nil
+}
+
+// writeCredentialsToTempFile writes credentials data to a temporary file and returns the filename
+func writeCredentialsToTempFile(data []byte) (string, error) {
+	f, err := os.CreateTemp("", "aws-shared-credentials")
```
gnufied, line 179: Will this work with `readOnlyFilesystem` field we are planning to add to operator pods? cc @dfajmon @jsafrane

gnufied: We will have to exclude `tmp` or mount `tmp` as emptydir for this to work once `readOnlyFileSystem` is enabled.

dfajmon: yes it's as you say, this would mean excluding the /tmp

pkg/driver/aws-ebs/aws_ebs.go
```diff
@@ -205,6 +205,11 @@ func GetAWSEBSOperatorControllerConfig(ctx context.Context, flavour generator.Cl
 		cfg.ExtraControlPlaneControllers = append(cfg.ExtraControlPlaneControllers, ctrl)
 	}
 
+	if flavour == generator.FlavourHyperShift {
```
gnufied, line 208: Does this needs to work only on hypershift?

pkg/driver/aws-ebs/aws_ebs_tags_retry_worker.go
gnufied: Yes this is an issue. Thanks for catching this.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
anirudhAgniRedhat: We have updated the implementation required for AWS session authentication. Now, the session is created using the role-arn data and will be re-authenticated if it expires. We will not need to create the temp file for authenticating the session.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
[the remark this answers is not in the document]
anirudhAgniRedhat: Implemented same!!

pkg/driver/aws-ebs/aws_ebs_tags_queue_worker.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+			c.removeVolumesFromQueueSet(volume.Name)
+			klog.Infof("Successfully updated PV annotations and tags for volume %s", volume.Name)
+		}
+	} else {
+		if len(item.pvNames) == 0 {
+			c.queue.Forget(item)
+			return
+		}
+		pv, err := c.getPersistentVolumeByName(item.pvNames[0])
```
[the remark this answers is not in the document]
dobsonj: If there are special cases for batch updates, you could use `len(item.pvNames) > 1` the same as `item.updateType == "batch"`. But it would be better to minimize those special cases and handle an array length of 1 the same as >1 as a general rule.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,510 @@
...
+		klog.Errorf("error updating volume %v tags: %v", pv.Name, updateErr)
+		c.queue.AddRateLimited(&pvUpdateQueueItem{
+			updateType: updateTypeIndividual,
+			pvNames:    convertPVsListToStringArray(pv),
+		})
+	}
+	var pvNames []string
+	for _, pv := range pvList {
+		pvNames = append(pvNames, pv.Name)
+	}
```
dobsonj, line 353: Can you collapse these two loops into a single loop, and build `pvNames` at the same time as requeuing the updates?

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,510 @@
...
+	awsEBSSecretName      = "ebs-cloud-credentials"
+	driverName            = "ebs.csi.aws.com"
+	tagHashAnnotationKey  = "ebs.openshift.io/volume-tags-hash"
+	batchSize             = 50
+
+	awsErrorVolumeNotFound = "InvalidVolume.NotFound"
+
+	defaultReSyncPeriod  = 30 * time.Minute
+	updateTypeBatch      = "batch"
+	updateTypeIndividual = "individual"
```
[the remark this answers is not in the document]
dobsonj: Please first check my other comments though, I wonder if we really need `updateType`.

pkg/driver/aws-ebs/aws_ebs_tags_controller.go
```diff
@@ -0,0 +1,505 @@
...
+func (c *EBSVolumeTagsController) getEC2Client(awsRegion string) (*ec2.EC2, error) {
+	if c.awsSession == nil || c.isSessionExpired() {
+		sess, err := c.createAWSSession(awsRegion)
+		if err != nil {
+			klog.Errorf("Failed to create AWS session: %v", err)
+			return nil, err
+		}
+		c.mutex.Lock()
+		c.awsSession = sess
+		c.mutex.Unlock()
```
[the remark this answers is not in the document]
dobsonj: Thanks a lot @anirudhAgniRedhat , this is a very nice improvement. I added a few follow-up comments, but you can mark this one resolved.

pkg/driver/aws-ebs/aws_ebs_tags_queue_worker.go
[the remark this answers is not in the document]
anirudhAgniRedhat: I was thinking of doing that in the same way but it makes the code a bit messy and a bit hard to understand.
putting updateType clarifies what kind of operation is expected to happen where In batch operation there can be a batch that will have only 1 volume. 
Putting some points down:

1. AWS errors are not very detailed. In a batch operation, they will return the first error message let's say in a batch operation: if the nth volume ID does not exist, it will return an error for that volume only maybe n+xth volume might give some different error message. which we are not sure of.
2. Better readability of code. 
3. Better Events can be published with error messages in case of batch and individual updates.
4. There can be cases where the user has again changed the infrastructure tags to previous tags, then we would not need to make more requests for a particular PV. IMHO including them within the same function will make it a bit confusing  

I just want to put my reasoning for this field else I am open to refactoring it in the same way as you suggested.
 WDYT?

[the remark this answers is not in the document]
dobsonj: Ok, thanks for explaining, I'm okay with keeping the update types as you have them.

One more idea to simplify this function a bit: would it be possible to refactor the code inside the if / else block into separate methods? So that `processVolumes` could just have the common code and then call the right method, instead of nesting the code inside the if / else?

```
switch item.updateType {
    case updateTypeBatch:
        c.processBatchVolumes(ctx, item, infra, ec2Client)
    case updateTypeIndividual:
        c.processIndividualVolumes(ctx, item, infra, ec2Client)
    default:
        panic("invalid update type: %v", item.updateType)
}
```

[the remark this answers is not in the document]
anirudhAgniRedhat: Thanks; I have added the same.

anirudhAgniRedhat:
Hey @dobsonj, do you feel anything else is needed for this change?
Else we can consider to merge this..

/retest

dobsonj:
> Hey @dobsonj, do you feel anything else is needed for this change? Else we can consider to merge this..

I think it looks good, thanks for all your work on this @anirudhAgniRedhat 

/lgtm
/approve

/cc @ropatil010 
for pre-merge testing + qe-approved label

/hold for https://github.com/openshift/cluster-storage-operator/pull/528 to merge first

anirudhAgniRedhat:
Hey @ropatil010 Have you completed the pre-merge testing? so we can close this.
Thanks

ropatil010:
Execution Results added here: 
1. https://issues.redhat.com/browse/STOR-2286?focusedId=26682882&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26682882 
2. https://issues.redhat.com/browse/STOR-2286?focusedId=26699747&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26699747 

Scenarios covered: 
1. OCP-79814 [AWS-EBS-CSI][HCP] Verify with single tag and can be update to new tags
2. OCP-79815 [AWS-EBS-CSI][HCP] Verify that 25+ tags are not supported
3. OCP-80179 [AWS-EBS-CSI][HCP] Verify with multiple tags => newKey3: newValue3, newKey4: newValue4
4. OCP-80180 [AWS-EBS-CSI][HCP] Verify that an existing tags can be removed newKey5: newValue5 => []
5. OCP-80181 [AWS-EBS-CSI][HCP] Verify that a tag key does not support 128+ characters and tag value does not support 256+ characters.

// Regression tests on existing scenarios: 
1. Daemonsets/Stateful sets.
2. Static Provisioning.
3. Local storage operator. 
4. Retain reclaimPolicy

Adding label based on the tested results. 
/label-qe-approved

ropatil010:
Adding label based on the tested results. 
/label qe-approved

anirudhAgniRedhat:
/unhold 

https://github.com/openshift/cluster-storage-operator/pull/528 merged

PR #299, opened 2024-10-15
NO-JIRA: Add OWNERS files for OpenStack-specific components/assets

stephenfin:
This will be helpful as we work on Hypershift integration (though I still expect we'll want/need Storage guidance and reviews).

Note that this still won't let us merge changes to `openshift/release` since the `OWNERS` for that are generated from the root `OWNERS` file. I'm okay with just adding us to the root `OWNERS` file here, on the assumption that we'll only modify Cinder/Manila changes, but I will wait feedback first.


stephenfin:
/test e2e-openstack-cinder-csi

Issue with must-gather

PR #298, opened 2024-10-15
OSASINFRA-3618: Add Cinder CSI support (part 3)

stephenfin:
This is part 3 of the process of merging the OpenStack Cinder CSI Driver operator into `csi-operator`.

This PR addresses the final step in the [migration guide][migration], namely removing code from `vendor/legacy` in this repository.

**Previous steps:**

- https://github.com/openshift/csi-operator/pull/278
- https://github.com/openshift/csi-operator/pull/281
- https://github.com/openshift/release/pull/56812
- https://github.com/openshift-eng/ocp-build-data/pull/5402
- https://github.com/openshift/openstack-cinder-csi-driver-operator/pull/180
- https://github.com/openshift/csi-operator/pull/275
- https://github.com/openshift-eng/ocp-build-data/pull/5425
- https://github.com/openshift/release/pull/57034

[migration]: https://github.com/openshift/enhancements/blob/master/enhancements/storage/csi-driver-operator-merge.md#process-of-moving-operators-to-csi-operator-monorepo

PR #278, opened 2024-09-17
OSASINFRA-3609: Add Cinder CSI support (part 1)

stephenfin:
This is part 1 of the process of merging the OpenStack Cinder CSI Driver operator into `csi-operator`.

This PR follows the first three steps in the [migration guide][migration], namely:

1. Avoid `.gitignore` related footguns

   This was not an issue since our `.gitignore` only ignored the binary if it was in the root of the repo

2. Move existing code into `csi-operator` repository

3. Add Dockerfiles for building images from new location

https://github.com/openshift/release/pull/56812 will switch our jobs over to build images from this new location.

**Commits:**

- `Squashed 'legacy/openstack-cinder-csi-driver-operator/' content from commit dcac15afd`
- `openstack-cinder: Add Dockerfile`

[migration]: https://github.com/openshift/enhancements/blob/master/enhancements/storage/csi-driver-operator-merge.md#process-of-moving-operators-to-csi-operator-monorepo

PR #302, opened 2024-10-17
OSASINFRA-3535: Add support for Hypershift to Cinder CSI

stephenfin:
In #278, #275 and #298, we worked to integrate Cinder CSI driver support here, into `csi-operator`. With that out of the way, we can now move onto working on Hypershift support.

The changes required here are mostly minimal: we need to ensure that we sync our config map contain configuration for the drivers to the correct namespace. We also need to ensure that we pass through the guest kubeconfig to the controllers. All of this has been done elsewhere and should not be surprising. The main change is to our syncer. I would suggest reading the commit messages for more context, but the tl;dr: is that we modify this so that (a) it's able to run without a source config map, and (b) it's able to dump the results in multiple locations.

**Related changes**

- https://github.com/openshift/cluster-storage-operator/pull/525
- https://github.com/openshift/hypershift/pull/4936

### DIFF
--- a/pkg/openstack-cinder/config/cloudinfo.go
+++ b/pkg/openstack-cinder/config/cloudinfo.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"os"
 	"sort"
 
 	"github.com/gophercloud/gophercloud/v2"
@@ -11,6 +12,8 @@ import (
 	azutils "github.com/gophercloud/utils/v2/openstack/compute/v2/availabilityzones"
 	"github.com/openshift/csi-operator/pkg/version"
 )
+
+const caFile = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
 
 // CloudInfo caches data fetched from the user's openstack cloud
 type CloudInfo struct {
@@ -80,6 +83,15 @@ func getCloudInfo() (*CloudInfo, error) {
 	// we represent version using commits since we don't tag releases
 	ua := gophercloud.UserAgent{}
 	ua.Prepend(fmt.Sprintf("csi-operator/%s", version.Get().GitCommit))
+
+	// our assets mount the CA file at a known path and we don't want to rely on other things
+	// setting the 'cafile' value in our clouds.yaml file to the same path, so we explicitly
+	// override this if a CA file is present
+	if _, err := os.Stat(caFile); err == nil {
+		// NOTE(stephenfin): gophercloud (or rather, the clientconfig package) doesn't currently
+		// provide the way to override configuration other than via environment variables
+		os.Setenv("OS_CACERT", caFile)
+	}
 
 	ci.clients.computeClient, err = clientconfig.NewServiceClient(context.TODO(), "compute", opts)
 	if err != nil {
--- a/pkg/openstack-cinder/config/configsync.go
+++ b/pkg/openstack-cinder/config/configsync.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
-	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -22,6 +21,7 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 
+	"github.com/openshift/csi-operator/pkg/clients"
 	"github.com/openshift/csi-operator/pkg/openstack-cinder/util"
 )
 
@@ -37,6 +37,8 @@ type ConfigSyncController struct {
 }
 
 const (
+	resyncInterval = 20 * time.Minute
+
 	sourceConfigKey   = "config"
 	targetConfigKey   = "cloud.conf"
 	enableTopologyKey = "enable_topology"
@@ -44,30 +46,22 @@ const (
 	infrastructureResourceName = "cluster"
 )
 
-func NewConfigSyncController(
-	operatorClient v1helpers.OperatorClient,
-	kubeClient kubernetes.Interface,
-	informers v1helpers.KubeInformersForNamespaces,
-	configInformers configinformers.SharedInformerFactory,
-	guestNamespace string,
-	resyncInterval time.Duration,
-	eventRecorder events.Recorder) factory.Controller {
-
+func NewConfigSyncController(c *clients.Clients) factory.Controller {
 	// Read configmap from user-managed namespace and save the translated one
 	// to the operator namespace
-	configMapInformer := informers.InformersFor(util.OpenShiftConfigNamespace)
-	c := &ConfigSyncController{
-		operatorClient:       operatorClient,
-		kubeClient:           kubeClient,
+	configMapInformer := c.KubeInformers.InformersFor(util.OpenShiftConfigNamespace)
+	controller := &ConfigSyncController{
+		operatorClient:       c.OperatorClient,
+		kubeClient:           c.KubeClient,
 		configMapLister:      configMapInformer.Core().V1().ConfigMaps().Lister(),
-		infrastructureLister: configInformers.Config().V1().Infrastructures().Lister(),
-		guestNamespace:       guestNamespace,
-		eventRecorder:        eventRecorder.WithComponentSuffix("ConfigSync"),
+		infrastructureLister: c.ConfigInformers.Config().V1().Infrastructures().Lister(),
+		guestNamespace:       c.GuestNamespace,
+		eventRecorder:        c.EventRecorder.WithComponentSuffix("ConfigSync"),
 	}
-	return factory.New().WithSync(c.sync).ResyncEvery(resyncInterval).WithSyncDegradedOnError(operatorClient).WithInformers(
-		operatorClient.Informer(),
+	return factory.New().WithSync(controller.sync).ResyncEvery(resyncInterval).WithSyncDegradedOnError(c.OperatorClient).WithInformers(
+		c.OperatorClient.Informer(),
 		configMapInformer.Core().V1().ConfigMaps().Informer(),
-	).ToController("ConfigSync", eventRecorder)
+	).ToController("ConfigSync", c.EventRecorder)
 }
 
 func (c *ConfigSyncController) Name() string {
--- a/pkg/openstack-cinder/config/configsync.go
+++ b/pkg/openstack-cinder/config/configsync.go
@@ -19,42 +19,44 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
-	"k8s.io/klog/v2"
 
 	"github.com/openshift/csi-operator/pkg/clients"
 	"github.com/openshift/csi-operator/pkg/openstack-cinder/util"
 )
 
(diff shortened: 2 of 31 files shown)


MaysaMacedo (changes requested):
The following check also needs to be removed to support Hypershift
https://github.com/openshift/csi-operator/pull/302/files#diff-9e4d5abc381c99fd5db77cd5b530830b31ceba8d952f57da3f095fa3bd7fcac8R120-R123

EmilienM:
/testwith openshift/hypershift/main/e2e-openstack-csi-cinder openshift/csi-operator#302 openshift/cluster-storage-operator#525

pkg/openstack-cinder/config/configsync.go
```diff
@@ -18,56 +16,107 @@ import (
...
+		// If we're in a hypershift deployment, we watch the clusters-${NAME} namespace
+		// (on the control plane cluster)...
+		configMapInformer = c.GetControlPlaneConfigMapInformer(c.ControlPlaneNamespace)
+
+		// ...and we don't currently support user-defined configuration so there's nothing to set
+		// for that
+
+		// ...but we now have two clusters to save to
+		configDestinations = []configDestination{
+			{c.ControlPlaneNamespace, util.CinderConfigName, c.ControlPlaneKubeClient},
```
MaysaMacedo, line 97: is it really needed in the control plane namespace?
I took a look on the assets generated and the DaemonSet that runs on the guest cluster seems to be the only one that needs it https://github.com/openshift/csi-operator/pull/302/files#diff-dece87039665ecf71f9003b48e52ce76685b59171f7c8e74ae0110f72b0526acR197.

stephenfin: The diff for `controller.yaml` is hidden by default so you need to expand that. When you do, you'll see it's used in the controller also.

https://github.com/openshift/csi-operator/pull/302/files#diff-8114e4393b7e73a8d7b1a309b1d0688968e54ffe7b996087cd410b780d386731R412

stephenfin:
/retest-required

/hold

I have tested this locally in a HCP context and the deployment succeeded. I'm not sure if I missed something in the standalone context so I'm deploying that now locally also. Let's recheck in the CI also to see if it was just a timeout.

stephenfin:
/unhold

Figured the issue out and reshuffled things around so that I was retrieving the `infrastructure` CR inside the `sync` loop. This should be good to go now.

pkg/openstack-cinder/config/configsync.go
```diff
@@ -42,22 +61,29 @@ const (
...
-		kubeClient:           c.KubeClient,
-		configMapLister:      configMapInformer.Lister(),
-		infrastructureLister: c.GetInfraInformer().Lister(),
-		namespace:            c.GuestNamespace,
-		eventRecorder:        c.EventRecorder.WithComponentSuffix("ConfigSync"),
+		operatorClient:         c.OperatorClient,
+		configMapLister:        configMapInformer.Lister(),
+		infrastructureLister:   c.GetInfraInformer().Lister(),
+		controlPlaneKubeClient: c.ControlPlaneKubeClient,
+		guestKubeClient:        c.ControlPlaneKubeClient,
```
mandre, line 82: Shouldn't it be `c.KubeClient` ?

mandre (commented):
I'll need to understand better how the new configmap syncer works in practice and look at CI results more closely. Can it cause issues on upgraded clusters?

pkg/openstack-cinder/config/configsync.go
stephenfin: It should, yes. I reworked this yesterday and must have fat-fingered this.

stephenfin:
> I'll need to understand better how the new configmap syncer works in practice and look at CI results more closely. Can it cause issues on upgraded clusters?

I called this out in the commit message, but the largest change in the config syncer is that we now only copy the `[BlockStorage]` section, meaning we ignore/drop the `[Global]`, `[LoadBalancer]`, and `[Metadata]` sections. I think this is definitely the correct thing to do for `[Global]` and `[LoadBalancer]`, since we don't want users providing auth info outside of the `clouds.yaml` file managed by cloud-credential-operator (among other things, doing so would make credential rotation more difficult) and load balancer configuration is irrelevant for the CSI drivers. The removal of support for configuring the `[Metadata]` section is a little more nuanced but again, I don't know if we want people tweaking this from [its default of `configDrive,metadataService`](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#metadata).

The only way I can envision this causing issue on an upgraded cluster is if either (a) users had *completely* overridden the auth info found in `clouds.yaml` with other credentials or (b) they had *partially* overridden the auth info found in `clouds.yaml`. Both seem unlikely. In the former case, the new user would need to be an admin or a set of application credentials for the same user or they wouldn't have permissions to attach/detach volumes from the instance. In the latter case, the user would also have needed to do this for CCM and every other operator or service that talks to OpenStack. As such, I don't think we need to worry about this.

That change aside, the other change is that we have obviously reworked things to support HCP. The crux of that is that while in standalone we source from one of two potential config maps and save the results to a config map in the operator namespace (`openshift-cluster-csi-drivers / cloud-conf`), in HCP we don't source from anywhere (by design) and save the results to *two* config maps, one in the management cluster (`clusters-${NAME} / cloud-conf`) and one in the guest cluster (`openshift-cluster-csi-drivers / cloud-conf`). This is done because both the controllers (running in the management cluster) and the drivers (running in the guest cluster) need a configuration file (even though I don't think the former actually uses the thing).

Let me know if any of that isn't clear

stephenfin:
/retest

At least `ci/prow/e2e-openstack` failed due to infra flake.

mandre:
@stephenfin we'll need to update openstack-test to match us dropping the global section of the cloud.conf: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_csi-operator/302/pull-ci-openshift-csi-operator-master-e2e-openstack/1857356418852589568

EmilienM:
/testwith openshift/hypershift/main/e2e-openstack-csi-cinder openshift/hypershift#4936

stephenfin:
/hold

While we wait on https://github.com/openshift/openstack-test/pull/202 to land and get a green CI :heavy_check_mark: for OpenStack jobs.

The other two failures are long-term failures that we need to address separately.

EmilienM:
/testwith openshift/hypershift/main/e2e-openstack-csi-cinder https://github.com/openshift/hypershift/pull/4936

smg247:
/testwith openshift/hypershift/main/e2e-openstack-csi-cinder https://github.com/openshift/hypershift/pull/4936

EmilienM:
@stephenfin This job was running in Mecha where a self signed certificate is being used and look at these logs (this looks like the issue I had in my environment, as well):

https://github.com/openshift/csi-operator/pull/302/checks?check_run_id=33062602418

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/multi-pr-openshift-csi-operator-302-openshift-hypershift-4936-e2e-openstack-csi-cinder/1857514674782736384/artifacts/e2e-openstack-csi-cinder/dump/artifacts/namespaces/clusters-9337ef3317244feda749/core/pods/logs/openstack-cinder-csi-driver-operator-687c8496d5-nzdzm-openstack-cinder-csi-driver-operator.log

stephenfin:
/testwith openshift/hypershift/main/e2e-openstack-csi-cinder https://github.com/openshift/hypershift/pull/4936

EmilienM:
Again: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/multi-pr-openshift-csi-operator-302-openshift-hypershift-4936-e2e-openstack-csi-cinder/1857734832038088704/artifacts/e2e-openstack-csi-cinder/dump/artifacts/namespaces/clusters-28ff1dd1f48d163bdbf2/core/pods/logs/openstack-cinder-csi-driver-operator-f7998dc98-wp9wt-openstack-cinder-csi-driver-operator.log

stephenfin:
/hold remove

https://github.com/openshift/openstack-test/pull/202 has landed.

stephenfin:
> Again: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/multi-pr-openshift-csi-operator-302-openshift-hypershift-4936-e2e-openstack-csi-cinder/1857734832038088704/artifacts/e2e-openstack-csi-cinder/dump/artifacts/namespaces/clusters-28ff1dd1f48d163bdbf2/core/pods/logs/openstack-cinder-csi-driver-operator-f7998dc98-wp9wt-openstack-cinder-csi-driver-operator.log

Indeed. https://github.com/openshift/cluster-storage-operator/pull/538 should fix that.

pkg/openstack-cinder/config/cloudinfo.go
```diff
@@ -81,6 +82,15 @@ func getCloudInfo() (*CloudInfo, error) {
 	ua := gophercloud.UserAgent{}
 	ua.Prepend(fmt.Sprintf("csi-operator/%s", version.Get().GitCommit))
 
+	// our assets mount the CA file at a known path and we don't want to rely on other things
+	// setting the 'cafile' value in our clouds.yaml file to the same path, so we explicitly
+	// override this if a CA file is present
+	if _, err := os.Stat("/etc/kubernetes/static-pod-resources/configmaps/cloud-config"); err == nil {
```
MaysaMacedo, line 88: it's worth creating a constant for the path `/etc/kubernetes/static-pod-resources/configmaps/cloud-config`

stephenfin: I'll move this if I need to respin.

pkg/openstack-cinder/config/cloudinfo.go
```diff
@@ -81,6 +82,15 @@ func getCloudInfo() (*CloudInfo, error) {
...
+		// NOTE(stephenfin): gophercloud (or rather, the clientconfig package) doesn't currently
+		// provide the way to override configuration other than via environment variables
+		os.Setenv("OS_CACERT", "/etc/kubernetes/static-pod-resources/configmaps/cloud-config")
```
mandre, line 91: Isn't `/etc/kubernetes/static-pod-resources/configmaps/cloud-config` a directory? If so we'll need to point to the file holding the ca-cert...

stephenfin:
/testwith openshift/hypershift/main/e2e-openstack-csi-cinder https://github.com/openshift/hypershift/pull/4936

EmilienM:
/testwith openshift/hypershift/main/e2e-openstack openshift/hypershift#4936
testing this one too as it's broken and we need it back sooner rather than later

EmilienM:
all changes here are only for OpenStack Cinder except a minor typo fix.

/approve

EmilienM:
/testwith openshift/hypershift/main/e2e-openstack https://github.com/openshift/hypershift/pull/4936

EmilienM:
/testwith openshift/hypershift/main/e2e-openstack-csi-cinder https://github.com/openshift/hypershift/pull/4936
infra error

EmilienM:
/testwith openshift/hypershift/main/e2e-openstack openshift/hypershift#4936

EmilienM:
let's merge what we have now and follow-up with bug fixes that are found in the HCP CI jobs.
/hold cancel

EmilienM:
> /retest-required

I think we're good. Waiting on QE approval and this is good to be shipped. We'll have PRs on top as bug fixes.

mandre:
> I think we're good. Waiting on QE approval and this is good to be shipped. We'll have PRs on top as bug fixes.

@rlobillo whenever you get a chance.

rlobillo:
/label qe-approved

e2e-openstack failure is unrelated

PR #303, opened 2024-10-18
NO-JIRA: azure-disk, azure-file sync code de-duplication

stephenfin:
Both drivers provides `syncCloudConfigGuest` and `syncCloudConfigStandalone` functions which were nearly identical. Identical save for the location that the resulting config maps were saved to: `Client.GuestNamespace` for the former and `Client.ControlPlaneNamespace` for the latter. However, the docs for `Clients.ControlPlaneNamespace` say:

> Namespace where to install CSI driver control plane. On standalone cluster, it's the same as the guest namespace,
> i.e. openshift-cluster-csi-drivers.

This makes the difference between the two functions meaningless and we can instead use `syncCloudConfigGuest` for both HCP and Standalone deployments. Do just this, renaming the method to drop the Guest suffix.


jsafrane:
sorry about a late review, it looks good. Can you please rebase?
/approve

stephenfin:
> sorry about a late review, it looks good. Can you please rebase?

Done :+1:

duanwei33:
The hypershift-e2e-aks and azure-disk/file-csi look good. We can monitor more CI later, but I think we can approve it now.
/verified by @duanwei33

PR #304, opened 2024-10-18
OCPBUGS-43562: Add missing proxy hook for `openstack-cinder` driver

stephenfin:
This is responsible for transforming the `config.openshift.io/inject-proxy: csi-driver` annotation in the asset into `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` configuration. This is not currently done by default so we need to add it. Do that now.

### DIFF
--- a/pkg/driver/openstack-cinder/openstack_cinder.go
+++ b/pkg/driver/openstack-cinder/openstack_cinder.go
@@ -131,7 +131,7 @@ func GetOpenStackCinderOperatorControllerConfig(ctx context.Context, flavour gen
 	cfg.AddDeploymentHookBuilders(c, withCABundleDeploymentHook, withConfigDeploymentHook)
 	cfg.DeploymentWatchedSecretNames = append(cfg.DeploymentWatchedSecretNames, cloudCredSecretName, metricsCertSecretName)
 
-	cfg.AddDaemonSetHookBuilders(c, withCABundleDaemonSetHook, withConfigDaemonSetHook)
+	cfg.AddDaemonSetHookBuilders(c, withCABundleDaemonSetHook, withClusterWideProxyDaemonSetHook, withConfigDaemonSetHook)
 	cfg.DaemonSetWatchedSecretNames = append(cfg.DaemonSetWatchedSecretNames, cloudCredSecretName)
 
 	configMapSyncer, err := createConfigMapSyncer(c)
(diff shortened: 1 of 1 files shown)


stephenfin:
/cc @gnufied
/cc @jsafrane

I think there's the same bug in the `aws-ebs` and `aws-efs` drivers. I see the `config.openshift.io/inject-proxy` annotation in both the `assets/overlays/aws-ebs/generated/standalone/node.yaml` and `assets/overlays/aws-efs/generated/standalone/node.yaml` files, yet there doesn't appear to be any hook to process those. Assuming I've groked that correctly, the second commit in the patch will address those. If I haven't, I'll drop that.

pkg/driver/common/operator/hooks.go
```diff
@@ -35,21 +36,29 @@ func NewDefaultOperatorControllerConfig(flavour generator.ClusterFlavour, c *cli
...
-// withClusterWideProxy adds the cluster-wide proxy config to the Deployment.
-func withClusterWideProxy(c *clients.Clients) (dc.DeploymentHookFunc, []factory.Informer) {
+// withClusterWideProxyDeploymentHook adds the cluster-wide proxy config to the Deployment.
+func withClusterWideProxyDeploymentHook(c *clients.Clients) (dc.DeploymentHookFunc, []factory.Informer) {
 	hook := csidrivercontrollerservicecontroller.WithObservedProxyDeploymentHook()
 	return hook, nil
 }
 
+// withClusterWideProxyHook adds the cluster-wide proxy config to the DaemonSet.
+func withClusterWideProxyDaemonSetHook(_ *clients.Clients) (csidrivernodeservicecontroller.DaemonSetHookFunc, []factory.Informer) {
```
gnufied, line 57: So, this hook is not needed for all CSI drivers because for EFS and EBS drivers, usually they don't talk to cloud on the node side at all (I could be wrong about this for EFS, will have to double check). So, that is why you don't see this applied for every driver in this package. 

IMO - I will keep this hook on per-driver basis for daemonset because not every driver needs this.

stephenfin: Okay. As I noted above though, [both the EBS and EFS drivers have the `config.openshift.io/inject-proxy` annotation](https://github.com/openshift/csi-operator/pull/304#issuecomment-2422834547). I can drop the second commit here, but should I replace it with one that drops those annotations since they're apparently not needed?

stephenfin: I've dropped this from this PR and will move to a separate PR.

stephenfin: PS: Separate PR is [here](https://github.com/openshift/csi-operator/pull/308).

PR #308, opened 2024-10-23
NO-JIRA: Remove `inject-proxy` annotations for aws-ebs, aws-efs node daemonsets

stephenfin:
There is no proxy hook configured for either driver making this a no-op. Remove it.

### DIFF
--- a/assets/overlays/aws-ebs/generated/hypershift/node.yaml
+++ b/assets/overlays/aws-ebs/generated/hypershift/node.yaml
@@ -14,7 +14,6 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   annotations:
-    config.openshift.io/inject-proxy: csi-driver
     config.openshift.io/inject-proxy-cabundle: csi-driver
   name: aws-ebs-csi-driver-node
   namespace: ${NODE_NAMESPACE}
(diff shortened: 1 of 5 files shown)


stephenfin:
@jsafrane @gnufied I've only removed the `config.openshift.io/inject-proxy: csi-driver` annotation here. However, I'm wondering if I should also remove the `config.openshift.io/inject-proxy-cabundle: csi-driver` annotation and the associated hooks. Based on feedback in https://github.com/openshift/csi-operator/pull/304, we're saying that the drivers don't need access to the cloud on the node and there is therefore no need for a proxy hook. If this is the case, what is the reason for [injecting CA bundles into the containers](https://github.com/openshift/csi-operator/blob/e71aab2165b230f4d3ecc301bc30e45cd9400a4c/pkg/driver/aws-ebs/aws_ebs.go#L314-L325)? Is that a no-op?

jsafrane:
/lgtm
/approve

/label px-approved
/label docs-approved
I leave it to QE to validate the AWS EBS and AWS EFS daemonsets do not currently talk to the cloud API.

Phaow:
/label qe-approved
The presubmit jobs `aws-efs-operator-e2e-extended` and `e2e-aws-csi-extended` already use the proxy configured cluster to execute the e2e tests, the results looks good. It seems azure file needs the proxy hook talk to the cloud API while both aws disk and file does not need.

stephenfin:
/retest-required

Both failures are clearly unrelated to the change.

PR #309, opened 2024-10-23
HOSTEDCP-2032: Set azure disk & file Secrets Store CSI Volume

bryan-cox:
Set the azure-disk and azure-file Secrets Store CSI volumes and volume mounts when the following environment variables are passed to the csi-operator pod - ARO_HCP_SECRET_PROVIDER_CLASS_FOR_FILE, ARO_HCP_SECRET_PROVIDER_CLASS_FOR_DISK.


jsafrane:
/lgtm
/approve
/label px-approved
/label docs-approved
/label qe-approved
will be tested in hypershift once everything merges.

PR #313, opened 2024-11-04
CFE-1132: EFS Access Point Tags Update DAY2

anirudhAgniRedhat:
This PR introduces a custom EFSAccessPointsTagController that monitors the OpenShift Infrastructure resource for changes in AWS ResourceTags. When tags are updated, the controller automatically fetches all AWS EFS-backed PersistentVolumes (PVs) in the cluster, retrieves their volume IDs, and updates the associated EFS Access Points tags in AWS.

Key Changes:

Monitors Infrastructure resource for AWS ResourceTags updates.
Directly fetches all PVs using the AWS EFS CSI driver (efs.csi.aws.com).
Updates AWS EFS AccessPoint tags by merging new and existing tags using the AWS SDK.

### DIFF
--- a/pkg/driver/aws-efs/aws_efs.go
+++ b/pkg/driver/aws-efs/aws_efs.go
@@ -107,12 +107,11 @@ func GetAWSEFSOperatorControllerConfig(ctx context.Context, flavour generator.Cl
 	cfg.DeploymentWatchedSecretNames = append(cfg.DeploymentWatchedSecretNames, cloudCredSecretName, metricsCertSecretName)
 	cfg.AddDaemonSetHookBuilders(c, withFIPSDaemonSetHook, withVolumeMetricsDaemonSetHook)
 	cfg.AddCredentialsRequestHook(stsCredentialsRequestHook)
-	if flavour == generator.FlavourHyperShift {
-		accessPointsTagController := NewEFSAccessPointTagsController(cfg.GetControllerName("EFSAccessPointTagsController"), c, c.EventRecorder)
-		cfg.ExtraControlPlaneControllers = append(cfg.ExtraControlPlaneControllers, accessPointsTagController)
-		cfg.DeploymentInformers = append(cfg.DeploymentInformers, c.KubeInformers.InformersFor("").Core().V1().PersistentVolumes().Informer())
-		cfg.DeploymentInformers = append(cfg.DeploymentInformers, c.KubeInformers.InformersFor(awsEFSSecretNamespace).Core().V1().Secrets().Informer())
-	}
+
+	accessPointsTagController := NewEFSAccessPointTagsController(cfg.GetControllerName("EFSAccessPointTagsController"), c, c.EventRecorder)
+	cfg.ExtraControlPlaneControllers = append(cfg.ExtraControlPlaneControllers, accessPointsTagController)
+	cfg.DeploymentInformers = append(cfg.DeploymentInformers, c.KubeInformers.InformersFor("").Core().V1().PersistentVolumes().Informer())
+	cfg.DeploymentInformers = append(cfg.DeploymentInformers, c.KubeInformers.InformersFor(awsEFSSecretNamespace).Core().V1().Secrets().Informer())
 
 	return cfg, nil
 }
--- a/pkg/driver/aws-efs/aws_efs.go
+++ b/pkg/driver/aws-efs/aws_efs.go
@@ -269,16 +269,26 @@ func withCustomTags(c *clients.Clients) (dc.DeploymentHookFunc, []factory.Inform
 			}
 			tagPairs = append(tagPairs, fmt.Sprintf("%s:%s", userTag.Key, userTag.Value))
 		}
+		tagPairs = append(tagPairs, fmt.Sprintf("kubernetes.io/cluster/%s:owned", infra.Status.InfrastructureName))
 
 		// Join the tag pairs with a space separator
 		tags := strings.Join(tagPairs, " ")
 		tagsArgument := fmt.Sprintf("--tags=%s", tags)
 
-		for _, container := range deployment.Spec.Template.Spec.Containers {
+		for i := range deployment.Spec.Template.Spec.Containers {
+			container := &deployment.Spec.Template.Spec.Containers[i]
 			if container.Name != "csi-driver" {
 				continue
 			}
-			container.Args = append(container.Args, tagsArgument)
+
+			// Remove any existing --tags argument
+			filteredArgs := make([]string, 0, len(container.Args))
+			for _, arg := range container.Args {
+				if !strings.HasPrefix(arg, "--tags=") {
+					filteredArgs = append(filteredArgs, arg)
+				}
+			}
+			container.Args = append(filteredArgs, tagsArgument)
 		}
 		return nil
 	}
--- a/pkg/driver/aws-efs/aws_efs.go
+++ b/pkg/driver/aws-efs/aws_efs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	opv1 "github.com/openshift/api/operator/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -102,10 +103,16 @@ func GetAWSEFSOperatorConfig() *config.OperatorConfig {
 // after a client connection + cluster flavour are established.
 func GetAWSEFSOperatorControllerConfig(ctx context.Context, flavour generator.ClusterFlavour, c *clients.Clients) (*config.OperatorControllerConfig, error) {
 	cfg := operator.NewDefaultOperatorControllerConfig(flavour, c, "AWSEFS")
-	cfg.AddDeploymentHookBuilders(c, withCABundleDeploymentHook, withFIPSDeploymentHook)
+	cfg.AddDeploymentHookBuilders(c, withCABundleDeploymentHook, withFIPSDeploymentHook, withCustomTags)
 	cfg.DeploymentWatchedSecretNames = append(cfg.DeploymentWatchedSecretNames, cloudCredSecretName, metricsCertSecretName)
 	cfg.AddDaemonSetHookBuilders(c, withFIPSDaemonSetHook, withVolumeMetricsDaemonSetHook)
 	cfg.AddCredentialsRequestHook(stsCredentialsRequestHook)
+	if flavour == generator.FlavourHyperShift {
+		accessPointsTagController := NewEFSAccessPointTagsController(cfg.GetControllerName("EFSAccessPointTagsController"), c, c.EventRecorder)
+		cfg.ExtraControlPlaneControllers = append(cfg.ExtraControlPlaneControllers, accessPointsTagController)
+		cfg.DeploymentInformers = append(cfg.DeploymentInformers, c.KubeInformers.InformersFor("").Core().V1().PersistentVolumes().Informer())
+		cfg.DeploymentInformers = append(cfg.DeploymentInformers, c.KubeInformers.InformersFor(awsEFSSecretNamespace).Core().V1().Secrets().Informer())
+	}
 
 	return cfg, nil
 }
@@ -234,4 +241,50 @@ func withVolumeMetricsDaemonSetHook(c *clients.Clients) (csidrivernodeservicecon
 	}
 	return hook, informers
 
+}
+
+// withCustomTags add tags from Infrastructure.Status.PlatformStatus.AWS.ResourceTags to the driver command line as
+// --tags=<key1>:<value1> <key2>:<value2>,...
+func withCustomTags(c *clients.Clients) (dc.DeploymentHookFunc, []factory.Informer) {
+	hook := func(spec *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		infraLister := c.GetInfraInformer().Lister()
+		infra, err := infraLister.Get(infrastructureName)
+		if err != nil {
+			return err
+		}
+		if infra.Status.PlatformStatus == nil || infra.Status.PlatformStatus.AWS == nil {
+			return nil
+		}
+
+		userTags := infra.Status.PlatformStatus.AWS.ResourceTags
+		if len(userTags) == 0 {
+			return nil
+		}
+
+		// Create a slice of formatted key:value pairs
+		tagPairs := make([]string, 0, len(userTags))
+		for _, userTag := range userTags {
+			// Skip tags with empty keys or values to avoid invalid formatting
+			if userTag.Key == "" || userTag.Value == "" {
+				continue
+			}
+			tagPairs = append(tagPairs, fmt.Sprintf("%s:%s", userTag.Key, userTag.Value))
+		}
+
+		// Join the tag pairs with a space separator
+		tags := strings.Join(tagPairs, " ")
+		tagsArgument := fmt.Sprintf("--tags=%s", tags)
+
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if container.Name != "csi-driver" {
+				continue
+			}
+			container.Args = append(container.Args, tagsArgument)
+		}
+		return nil
+	}
+	informers := []factory.Informer{
+		c.GetInfraInformer().Informer(),
+	}
+	return hook, informers
 }
--- a/pkg/driver/aws-efs/aws_efs_tags_controller.go
+++ b/pkg/driver/aws-efs/aws_efs_tags_controller.go
@@ -2,20 +2,16 @@ package aws_efs
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/base64"
-	"encoding/hex"
-	"encoding/json"
 	"fmt"
-	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
-	"github.com/aws/aws-sdk-go/service/sts"
-	"golang.org/x/time/rate"
-	"gopkg.in/ini.v1"
 	"os"
 	"sort"
 	"strings"
 	"sync"
 	"time"
+
+	"crypto/sha256"
+	"encoding/hex"
+	"gopkg.in/ini.v1"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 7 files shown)


anirudhAgniRedhat:
@jsafrane this PR adds support for updating tags on efs resources. 
Unlike the EBS Volumes, I didn't find way to update the tags in batch API calls. 

Here I am annotating the volumes with the sorted hash as discussed in [this thread.](https://github.com/openshift/csi-operator/pull/297#discussion_r1811287907)
So we can retry the tags the resources again in every re-sync period that is 20 mins and are reporting the failures as a warning events. 

Can you please review these changes as well.

/cc @jsafrane @TrilokGeer

gnufied:
If this is the new PR, can you close older PR?

pkg/driver/aws-efs/aws_efs.go
```diff
@@ -243,3 +248,50 @@ func withVolumeMetricsDaemonSetHook(c *clients.Clients) (csidrivernodeservicecon
...
 
 }
+
+// withCustomTags add tags from Infrastructure.Status.PlatformStatus.AWS.ResourceTags to the driver command line as
+// --tags=<key1>:<value1> <key2>:<value2>,...
+func withCustomTags(c *clients.Clients) (dc.DeploymentHookFunc, []factory.Informer) {
+	hook := func(spec *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		infraLister := c.GetInfraInformer().Lister()
+		infra, err := infraLister.Get(infrastructureName)
+		if err != nil {
```
gnufied, line 258: I would move this code to a different function and perform this computation only if container in question is `csi-driver`

pkg/driver/aws-efs/aws_efs_tags_controller.go
```diff
@@ -0,0 +1,413 @@
...
+	var infraRegion = ""
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil {
+		infraRegion = infra.Status.PlatformStatus.AWS.Region
+	}
+	efsClient, err := c.getEFSClient(ctx, infraRegion)
+	if err != nil {
+		return err
+	}
+
+	limiter := rate.NewLimiter(rate.Limit(awsTagsRequestRateLimit), awsTagsRequestBurstSize)
```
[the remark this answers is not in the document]
anirudhAgniRedhat: I have improved the implementation of the rate limiter here.

I am  using the limiter to not to hit more than 10 req/sec as listed in AWS [Service Quotas](https://docs.aws.amazon.com/servicequotas/latest/userguide/reference_limits.html)

pkg/driver/aws-efs/aws_efs_tags_controller.go
```diff
@@ -0,0 +1,413 @@
...
+		Tags:       tags,
+	})
+	if err != nil {
+		klog.Errorf("Error updating tags for PV %s: %v", pv.Spec.CSI.VolumeHandle, err)
+		return err
+	}
+	return nil
+}
+
+func (c *EFSAccessPointTagsController) updateVolumeWithRetry(ctx context.Context, pv *v1.PersistentVolume) error {
```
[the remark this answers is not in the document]
anirudhAgniRedhat: I see an issue with using the syncContext Queue as it will again call the sync function if anything is added to the queue.

I have new retry worker irrespective of the sync function logic and will solve our retry issues.

pkg/driver/aws-efs/aws_efs_tags_controller.go
```diff
@@ -0,0 +1,478 @@
...
+func (c *EFSAccessPointTagsController) getEFSClient(awsRegion string) (*efs.EFS, error) {
+	if c.awsSession == nil || c.isSessionExpired() {
+		sess, err := c.createAWSSession(awsRegion)
+		if err != nil {
+			klog.Errorf("Failed to create AWS session: %v", err)
+			return nil, err
+		}
+		c.mutex.Lock()
+		c.awsSession = sess
+		c.mutex.Unlock()
```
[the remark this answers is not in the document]
dobsonj: Actually I see now it's also called from a separate goroutine in aws_efs_tags_retry_worker.go too. I think it would be preferable to limit this to a single Sync loop, similar to what @TrilokGeer mentioned here:
https://github.com/openshift/csi-operator/pull/297#discussion_r1886633345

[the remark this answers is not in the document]
dobsonj: It also seems wrong to lock `c.awsSession` when you write it but not when you read it. That problem goes away if it's single-threaded.

pkg/driver/aws-efs/aws_efs_tags_controller.go
```diff
@@ -0,0 +1,478 @@
...
+			klog.Errorf("Error updating PV annotations for volume %s: %v", volume.Name, err)
+			c.addToFailedQueue(volume.Name)
+			continue
+		}
+		klog.Infof("Successfully updated PV annotations and access points tags for volume %s", volume.Name)
+	}
+	return nil
+}
+
+// updateEFSTags updates the tags of an AWS EFS Access Points
```
dobsonj, line 324:
```suggestion
// updateEFSAccessPointTags updates the tags of an AWS EFS Access Points
```

pkg/driver/aws-efs/aws_efs_tags_controller.go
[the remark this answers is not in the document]
anirudhAgniRedhat: I have improved the usage of mutex now it should be thread-safe. I have locked the session to be used again from any other thread before it is updated and unlocked.
Also AFAIU we cannot limit the update calls only from the sync loop because the Sync loop will only called to update the newer resources or newer updates from the infrastructure resource if there are any. we will need to update the failed volumes from the retry queue worker. 

Alternatively, I can propose other ways:

- Use the different session variable for retry queue worker. (It will create a duplicay though for sessions)
- Create a new worker just to call the AWS APIS taking []volumes as an argument and leverage the channels to update tags on the volumes from the single worker. but it would add a bit more complexity to the code.

I am open to any other suggestions as well.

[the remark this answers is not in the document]
dobsonj: Ok, I see the problem with completely serializing the Sync loop now, thanks for that context.

Now I have a different concern with holding the mutex for the duration of `getEFSClient()`. It is safe in the sense that it covers both read and write of `c.awsSession`, but now the mutex is held during two blocking calls to external services (one to read the secret from the k8s API server, another to authenticate and create the AWS session). It's generally a bad idea to hold a mutex during blocking network calls like this. If one thread is waiting to receive a response from AWS for example (potentially until timeout), then the other thread will be stuck waiting on the mutex. It behaves as if it's single-threaded, but with the added complexity of coordinating multiple threads.

I think you're onto something with this idea, but you may not need to use channels:

> Create a new worker just to call the AWS APIS taking []volumes as an argument and leverage the channels to update tags on the volumes from the single worker. but it would add a bit more complexity to the code.

What if the main Sync loop only processes the Infrastructure object and finds the PV's to be updated. Instead of calling `getEFSClient`, `updateEFSAccessPointTags`, and `updateVolume` directly, it just adds the volume.Name to the workqueue, and then the Sync loop is done. The retry worker responds by getting the queue entry and making the calls to update the EFS tags and the PV.

If you separate the duties of each thread like that:
* You still need a mutex for the work queue, but not for the awsSession, since only one worker uses awsSession.
* You can eliminate some code in `fetchPVsAndUpdateTags`, because it just adds the volume to the work queue, instead of trying to update the tags and only adding it to the work queue if the update failed.
* You can probably remove `c.rateLimiter` and the calls to `c.rateLimiter.Wait(ctx)` since the work queue already has a rate limit and you no longer have to limit the total number of AWS updates across two threads.

anirudhAgniRedhat:
Hey @dobsonj I have added updated the same as we did for the EBS operator.
Thanks for the suggestion.

Putting down the changes in the last commit.

1. Using a single worker just to update the tags by AWS API calls.
2. Passing PV name in queue which will further update and retry with backoff period.
3. Creating a new session on every tag update request.
4. Events on failures
5. Removed mutex locking for session.
6. Added Unit tests
7. Updated aws session auth method
8. Using [infra.status.controlPlaneTopology == External](https://redhat-external.slack.com/archives/C01C8502FMM/p1739369738688739?thread_ts=1739256240.732729&cid=C01C8502FMM) for validating the cluster is a hypershift cluster. As it is better to use than label. I have linked the slack conversation for same.


Things to consider here.

- The auth will need a web_identity token to get the authentication. Either it will be coming from the deployments or we have to use the https://github.com/openshift/aws-pod-identity-webhook 
  - If we would need to use the https://github.com/openshift/aws-pod-identity-webhook we have to add the following annotation on the serviceAccount.
    - eks.amazonaws.com/role-arn="<aws-arn-id>"
    - eks.amazonaws.com/audience="sts.amazonaws.com"
  -  This is required for [aws-pod-identity-webhook](https://github.com/openshift/aws-pod-identity-webhook) running on the cluster to correctly assign AWS roles to the pod(s) used by operator.
- Without this we will see the auth issues

We can discuss the alternative if you have any.

pkg/driver/aws-efs/aws_efs_tags_controller.go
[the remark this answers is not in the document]
anirudhAgniRedhat: Hey @dobsonj I have added updated the same as we did for the EBS operator.
Thanks for the suggestion.

Putting down the changes in the last commit.

1. Using a single worker just to update the tags by AWS API calls.
2. Passing PV name in queue which will further update and retry with backoff period.
3. Creating a new session on every tag update request.
4. Events on failures
5. Removed mutex locking for session.
6. Added Unit tests
7. Updated aws session auth method
8. Using [infra.status.controlPlaneTopology == External](https://redhat-external.slack.com/archives/C01C8502FMM/p1739369738688739?thread_ts=1739256240.732729&cid=C01C8502FMM) for validating the cluster is a hypershift cluster. As it is better to use than label. I have linked the slack conversation for same.


Things to consider here.

- The auth will need a web_identity token to get the authentication. Either it will be coming from the deployments or we have to use the https://github.com/openshift/aws-pod-identity-webhook 
  - If we would need to use the https://github.com/openshift/aws-pod-identity-webhook we have to add the following annotation on the serviceAccount.
    - eks.amazonaws.com/role-arn="<aws-arn-id>"
    - eks.amazonaws.com/audience="sts.amazonaws.com"
  -  This is required for [aws-pod-identity-webhook](https://github.com/openshift/aws-pod-identity-webhook) running on the cluster to correctly assign AWS roles to the pod(s) used by operator.
- Without this we will see the auth issues

We can discuss the alternative if you have any.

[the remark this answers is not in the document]
anirudhAgniRedhat: I figured out another way to do this as we are doing in the aws-efs-csi-driver-controller deployment.
We would need update the csv for this. Not sure if this would break the standalone cluster!
We can Mount the volume, secrets and bound-sa token to get it authenticated.
```
volumeMounts:
  - mountPath: /var/run/aws/auth
    name: aws-credentials
    readOnly: true
  - mountPath: /var/run/secrets/openshift/serviceaccount
    name: bound-sa-token
    readOnly: true

volumes:
  - name: aws-credentials
    secret:
      items:
        - key: credentials
          path: credentials
      secretName: aws-efs-cloud-credentials
  - name: bound-sa-token
    projected:
      sources:
        - serviceAccountToken:
            audience: openshift
            expirationSeconds: 3600
            path: token
            
```


Also would need to add 2 new env vars
```
- name: AWS_CONFIG_FILE
  value: /var/run/aws/auth/credentials
- name: AWS_SDK_LOAD_CONFIG
  value: '1'
```               

What would you suggest here?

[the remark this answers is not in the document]
anirudhAgniRedhat: I tested the changes on standalone and hyper-shift  cluster
For a fresh installation, we would need to make sure the secrets are created before the installation for the operator happens or the user has to apply which will create a credential request and auth secret. 
```
apiVersion: operator.openshift.io/v1
kind: ClusterCSIDriver
metadata:
    name: efs.csi.aws.com
spec:
  logLevel: Normal
  managementState: Managed
  operatorLogLevel: Trace
```
before the operator installation happens.

pkg/driver/aws-efs/aws_efs_tags_controller.go
```diff
@@ -0,0 +1,391 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// isHypershiftCluster validates the cluster is hypershift cluster or not
+func isHypershiftCluster(infra *configv1.Infrastructure) bool {
+	if infra.Status.ControlPlaneTopology == configv1.ExternalTopologyMode {
```
dobsonj, line 115: AFAIK, `ExternalTopologyMode` is not a reliable indicator of a hypershift cluster because it may be used for other flavors of openshift as well. IBM ROKS [for example](https://github.com/openshift/cluster-storage-operator/blob/30a471d023eb0e58d6b9735fadb7711d3bdb5b2a/pkg/operator/csidriveroperator/csioperatorclient/ibm-vpc-block.go#L22-L24) uses it (but of course aws-ebs is not running on IBM ROKS).

I think it would be better to check the `hypershift.openshift.io/managed: "true"` label if the goal is to enable it only on hypershift deployments.

pkg/driver/aws-efs/aws_efs_tags_controller.go
```diff
@@ -0,0 +1,395 @@
...
+	eventRecorder events.Recorder
+	queue         workqueue.TypedRateLimitingInterface[string]
+	queueSet      map[string]struct{} // A set to track added volume names
+	mutex         sync.Mutex
+}
+
+// TokenClaims represents the JWT claims
+type TokenClaims struct {
+	Exp int64 `json:"exp"` // Expiry timestamp
+}
```
dobsonj, line 58: Nit: can this be removed? I don't see it referenced anymore.

anirudhAgniRedhat: Removed, Missed this while refactoring the code.

dobsonj:
Thanks for all your work on this @anirudhAgniRedhat

/lgtm
/approve

/cc @ropatil010
for pre-merge testing + qe-approved label

dobsonj:
assuming the presubmit jobs pass of course :)
/retest

pkg/driver/aws-efs/aws_efs.go
```diff
@@ -235,3 +242,50 @@ func withVolumeMetricsDaemonSetHook(c *clients.Clients) (csidrivernodeservicecon
...
+			}
+			tagPairs = append(tagPairs, fmt.Sprintf("%s:%s", userTag.Key, userTag.Value))
+		}
+
+		// Join the tag pairs with a space separator
+		tags := strings.Join(tagPairs, " ")
+		tagsArgument := fmt.Sprintf("--tags=%s", tags)
+
+		for i := range deployment.Spec.Template.Spec.Containers {
+			container := &deployment.Spec.Template.Spec.Containers[i]
```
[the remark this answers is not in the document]
anirudhAgniRedhat: The  loop using `for _, container := range ...` creates a copy of the container instead of modifying the actual object in the slice. This results in no changes being applied to the deployment.

Moved it to same way

pkg/driver/aws-efs/aws_efs.go
```diff
@@ -235,3 +241,50 @@ func withVolumeMetricsDaemonSetHook(c *clients.Clients) (csidrivernodeservicecon
...
+			if container.Name != "csi-driver" {
+				continue
+			}
+			container.Args = append(container.Args, tagsArgument)
```
Phaow, line 282: We should not append the args directly since the `--tags=` already has default value at https://github.com/openshift/csi-operator/blob/main/assets/overlays/aws-efs/generated/standalone/controller.yaml#L53, otherwise the container args will have 2 `--tags=` as unexpected(I guess the second one will be override the first one), it is a bit different from ebs which we use a different arg `--extra-tags=` for the customTags. We'd better get the origin `--tags=` value and append the extra tags to origin value. BTW I saw the https://issues.redhat.com/browse/OCPSTRAT-787 is only used for hcp, right?
```console
#  2 "--tags=" exist
$ och get deploy aws-efs-csi-driver-controller -o jsonpath="{.spec.template.spec.containers[?(@.name=='csi-driver')].args}"
["--endpoint=$(CSI_ENDPOINT)","--logtostderr","--tags=kubernetes.io/cluster/hypershift-ci-338091:owned","--delete-access-point-root-dir=true","--v=8","--tags=newKey5:newValue5"]
```

cc to @dobsonj @gnufied

pkg/driver/aws-efs/aws_efs.go
```diff
@@ -102,11 +103,16 @@ func GetAWSEFSOperatorConfig() *config.OperatorConfig {
 // after a client connection + cluster flavour are established.
 func GetAWSEFSOperatorControllerConfig(ctx context.Context, flavour generator.ClusterFlavour, c *clients.Clients) (*config.OperatorControllerConfig, error) {
 	cfg := operator.NewDefaultOperatorControllerConfig(flavour, c, "AWSEFS")
-	cfg.AddDeploymentHookBuilders(c, withCABundleDeploymentHook, withFIPSDeploymentHook)
+	cfg.AddDeploymentHookBuilders(c, withCABundleDeploymentHook, withFIPSDeploymentHook, withCustomTags)
```
Phaow, line 106:
```suggestion
	cfg.AddDeploymentHookBuilders(c, withCABundleDeploymentHook, withFIPSDeploymentHook)
	if flavour == generator.FlavourHyperShift {
		// hcp-only hooks
		cfg.AddDeploymentHookBuilders(c, withCustomTags)
	}
```
If the `withCustomTags` is used for hcp only, better to only add the hook for hcp.
cc to @dobsonj @gnufied

anirudhAgniRedhat: Hey @Phaow `flavour == generator.FlavourHyperShift` would not work for the efs operator as there is only a single manifest( [Slack Thread for reference](https://redhat-internal.slack.com/archives/C057L1GKYSK/p1738941712139239?thread_ts=1738081635.004619&cid=C057L1GKYSK) ). Also the EBS operator also [supports this](https://github.com/openshift/csi-operator/blob/main/pkg/driver/aws-ebs/aws_ebs.go#L378) .

anirudhAgniRedhat: For efs operator we check the operator is a hyprshift cluster or not based on the `hypershift.openshift.io/managed: "true"` label and it seems like a bad idea to make requests to the API server during operator initialization. Instead, we always let the controller start, and then Sync in aws_efs_tags_controller.go can return if the infastructure does not have that label.

Phaow: Good to know, thanks for the clarify!

pkg/driver/aws-efs/aws_efs.go
anirudhAgniRedhat: I have updated the implementation, this will ensure only correct tags are being passed for the tags update, including the tag by default.

ropatil010:
Adding label based on the test results included here: https://issues.redhat.com/browse/STOR-2286?focusedId=26729430&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26729430 
/label qe-approved
/retest

PR #316, opened 2024-11-06
OSASINFRA-3654: Include OpenStack test manifest to test/e2e

MaysaMacedo:
This commit includes the test manifest from both cinder and manila to the test/e2e folder. This is one of the required post migration steps.


MaysaMacedo:
The files were taken from:
https://github.com/openshift/openstack-cinder-csi-driver-operator/blob/master/test/e2e/manifest.yaml
https://github.com/openshift/csi-driver-manila-operator/blob/master/test/e2e/manifest.yaml

I noticed other operators run an additional [test](https://github.com/openshift/csi-operator/blob/master/test/e2e/aws-ebs/ocp-manifest.yaml) `LUNStressTest`, but I don't think we ever ran this, so I will keep it out for now.

PR #321, opened 2024-11-20
OSASINFRA-3535: openstack-cinder: Watch management cluster for cloud-provider config

stephenfin:
We (unfortunately) need to extract our CA cert from the cloud-provider configuration. In a Hypershift context, this lives in the control plane cluster and has a different name. Correct both the cluster and name used and remove a check for missing config maps: this should *always* exist in a healthy cluster, given `cloud-provider-openstack` has a hard dependency on it.

### DIFF
--- a/pkg/openstack-cinder/config/configsync.go
+++ b/pkg/openstack-cinder/config/configsync.go
@@ -160,7 +160,7 @@ func (c *ConfigSyncController) sync(ctx context.Context, syncCtx factory.SyncCon
 	}
 
 	// ...and we watch the clusters-${NAME} namespace (on the control plane cluster)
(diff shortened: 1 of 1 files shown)


stephenfin:
I've tested this in a hypershift cluster and it works as expected. However, I'm maintaining the hold on this while we wait for the CI jobs to come back. iirc, there were some issues with `Clients.ControlPlaneKubeInformers` being `nil` in a standalone deployment. If that proves to be the case, we'll need to tweak this further. Or fix that issue (IMO, in a standalone deployment `ControlPlaneKubeInformers` should be identical to `KubeInformers`).

EmilienM:
@stephenfin do we need an equivalent in Manila?

PR #295, opened 2024-10-09
NO-JIRA: ignore cmd/generator/main.go in .snyk

dobsonj:
https://github.com/openshift/csi-operator/pull/287 introduced some new noise in the ci/prow/security job:

```
 ✗ [Medium] Path Traversal
   ID: d4eae1cf-1400-4b06-a906-bc189fed8faf 
   Path: cmd/generator/main.go, line 39 
   Info: Unsanitized input from a CLI argument flows into os.RemoveAll, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to delete arbitrary files.
```

cmd/generator/main.go is just a tool to generate static assets, not something an attacker could leverage to delete a file they couldn't otherwise delete.

/cc @openshift/storage

PR #323, opened 2024-11-21
NO-JIRA: snyk: Ignore irrelevant path traversal lint

stephenfin:


stephenfin:
/close

https://github.com/openshift/csi-operator/pull/295 predates this and is simpler.

PR #325, opened 2024-11-21
STOR-2017: Bump library-go and API

gnufied:


dobsonj:
/lgtm

But let's check the payload job results before adding the other labels
/payload-job periodic-ci-openshift-release-master-ci-4.18-upgrade-from-stable-4.17-e2e-aws-ovn-upgrade

cmd/aws-efs-csi-driver-operator/main.go
```diff
@@ -4,10 +4,11 @@ import (
 	"context"
 	"os"
 
-	"github.com/openshift/csi-operator/pkg/driver/aws-efs"
+	aws_efs "github.com/openshift/csi-operator/pkg/driver/aws-efs"
```
mpatlasov, line 7: Looks as NOOP change (make succeeds with or without this change)

gnufied: fair point. But lets fix this in a follow up, I do not want to run entire battery of tests for this change.

dobsonj:
one timeout... and we need to break out of this infinite loop of retesting
/override ci/prow/e2e-azure-csi

PR #327, opened 2024-11-22
OSASINFRA-3626: Resize support with OpenStack Manila CSI Driver

mandre:
Volume expansion is a mandatory feature supported
within back end drivers in OpenStack Manila. To
enable this with the Manila CSI driver, the operator needs to specify the external resizer sidecar along with its appropriate RBAC, and set "allowVolumeExpansion" to True. Volume expansion for NAS shared file systems is a controller-only operation; there's no node plugin operations necessary unlike block storage.


mandre:
/close
Closing in favor of #322

PR #329, opened 2024-11-22
NO-JIRA: build only needed cmd in container image

sanchezl:


jsafrane:
/retitle NO-JIRA: build only needed cmd in container image
/retest-required
/lgtm
/approve

This is not an user visible change
/label px-approved
/label docs-approved
/label qe-approved
CI should be enough to test this

PR #332, opened 2024-11-26
OCPBUGS-45073: Support HCP labels

muraee:
HCP is adding a new API field to allow users creating HCP clusters to set custom labels on all HCP pods https://github.com/openshift/hypershift/pull/5114. The CSI needs to integrate this feature in the same way it integrated the NodeSelector feature for HCP.


pkg/driver/common/operator/hooks.go
```diff
@@ -141,6 +141,40 @@ func getHostedControlPlaneTolerations(hostedControlPlaneLister hypev1beta1lister
...
+// withHyperShiftLabels sets Deployment labels on a HyperShift hosted control-plane.
+func withHyperShiftLabels(c *clients.Clients) (dc.DeploymentHookFunc, []factory.Informer) {
+	hook := func(_ *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		labels, err := getHostedControlLabels(
+			c.ControlPlaneHypeInformer.Hypershift().V1beta1().HostedControlPlanes().Lister(),
+			c.ControlPlaneNamespace)
+		if err != nil {
+			return err
+		}
+		deployment.Spec.Template.Labels = labels
```
jsafrane, line 153: This will replace all existing labels. The new labels should be *added*, so the deployment's labelSelector still works. In addition, existing labels should not be replaced with new values, again because of the deployment's labelSelector.

jsafrane:
In addition, you need to modify cluster-storage-operator to run the CSI driver operators themselves with the right labels, probably around [here](https://github.com/openshift/cluster-storage-operator/blob/094fc6c752dd95d8627def20981a3833415f7ca2/pkg/operator/csidriveroperator/hypershift_deployment_controller.go#L90).

muraee:
@jsafrane thanks for the hint about the needed change in cluster-storage-operator, I will open a PR for that.

pkg/driver/common/operator/hooks.go
```diff
@@ -141,6 +141,49 @@ func getHostedControlPlaneTolerations(hostedControlPlaneLister hypev1beta1lister
...
+		if err != nil {
+			return err
+		}
+
+		if deployment.Spec.Template.Labels == nil {
+			deployment.Spec.Template.Labels = map[string]string{}
+		}
+
+		for key, value := range labels {
+			// don't replace existing labels as theyt are used in the deployment's labelSelector.
```
jsafrane, line 159: typo: `theyt`

pkg/driver/common/operator/hooks_test.go
```diff
@@ -146,6 +146,55 @@ func Test_WithHyperShiftNodeSelector(t *testing.T) {
...
+		name           string
+		hcp            *hypev1beta1api.HostedControlPlane
+		expectedLabels map[string]string
+	}{
+		{
+			name:           "no labels",
+			hcp:            getTestHostedControlPlane("hcp_no_labels.yaml"),
+			expectedLabels: nil,
+		},
+		{
```
jsafrane, line 160: Please add a test that existing labels are not replaced

muraee: added

PR #344, opened 2024-12-09
OSASINFRA-3683: openstack-manila: Remove --nodeid parameter

stephenfin:
This option is deprecated as the driver can now auto-configure this [1]. We can also remove some unnecessary quoting.

Note that we don't remove the option from the NFS driver as that is a separate binary that doesn't support this auto-configuration.

[1] https://github.com/kubernetes/cloud-provider-openstack/pull/2734

Signed-off-by: Stephen Finucane <stephenfin@redhat.com>

### DIFF
--- a/assets/overlays/openstack-manila/base/node_nfs.yaml
+++ b/assets/overlays/openstack-manila/base/node_nfs.yaml
@@ -38,9 +38,9 @@ spec:
               cpu: 10m
           args:
             - --v=${LOG_LEVEL}
-            - "--nodeid=$(NODE_ID)"
-            - "--endpoint=unix://plugin/csi.sock"
-            - "--mount-permissions=0777"
+            - --nodeid=$(NODE_ID)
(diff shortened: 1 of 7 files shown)


stephenfin:
/hold

The code only just merged upstream, so this can't merge until after the next rebase of https://github.com/openshift/cloud-provider-openstack

assets/overlays/openstack-manila/base/node_nfs.yaml
```diff
@@ -38,9 +38,9 @@ spec:
               cpu: 10m
           args:
             - --v=${LOG_LEVEL}
-            - "--nodeid=$(NODE_ID)"
-            - "--endpoint=unix://plugin/csi.sock"
-            - "--mount-permissions=0777"
+            - --nodeid=$(NODE_ID)
```
MaysaMacedo, line 41: Why couldn't this also be removed from the NFS plugin?

PR #353, opened 2025-01-16
OCPBUGS-48507: openstack-manila: update PodDisruptionBudget name

MaysaMacedo:
In previous releases the PodDisruptionBudget is named openstack-manila-csi-controllerplugin-pdb instead of manila-csi-driver-controller-pdb, which is the name currently generated by csi-operator. This prevents upgrades from being successful. To fix the issue the asset is renamed with the same name as used in previous releases.


EmilienM:
/approve
note: hypershift-e2e-openstack-aws-csi-manila is not required but I want to make sure it passes as well.

MaysaMacedo:
/hold

I tested upgrades locally and it passed, but I still would like to have CI passing first.

mandre (commented):
/lgtm

PR #364, opened 2025-03-12
OCPBUGS-50493: Use 127.0.0.1 for healtz http-endpoints

mpatlasov:
Those ports are used only for internal communication inside CSI driver pods. No need to expose them for the world.

### DIFF
--- a/pkg/driver/common/generator/base_assets.go
+++ b/pkg/driver/common/generator/base_assets.go
@@ -5,12 +5,13 @@ import (
 )
 
 const (
-	ProvisionerAssetName         = "common/sidecars/provisioner.yaml"
-	AttacherAssetName            = "common/sidecars/attacher.yaml"
-	SnapshotterAssetName         = "common/sidecars/snapshotter.yaml"
-	ResizerAssetName             = "common/sidecars/resizer.yaml"
-	LivenessProbeAssetName       = "common/sidecars/livenessprobe.yaml"
-	NodeDriverRegistrarAssetName = "common/sidecars/node_driver_registrar.yaml"
+	ProvisionerAssetName              = "common/sidecars/provisioner.yaml"
+	AttacherAssetName                 = "common/sidecars/attacher.yaml"
+	SnapshotterAssetName              = "common/sidecars/snapshotter.yaml"
+	ResizerAssetName                  = "common/sidecars/resizer.yaml"
+	PodNetworkLivenessProbeAssetName  = "common/sidecars/pod_network_livenessprobe.yaml"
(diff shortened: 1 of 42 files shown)


pkg/driver/common/generator/base_assets.go
```diff
@@ -111,9 +112,15 @@ var (
...
-	DefaultLivenessProbe = generator.SidecarConfig{
-		TemplateAssetName: LivenessProbeAssetName,
+	// DefaultLivenessProbeDeployment is definition of the default livenessprobe sidecar for Deployment.
+	DefaultLivenessProbeDeployment = generator.SidecarConfig{
+		TemplateAssetName: LivenessProbeDeploymentAssetName,
+		ExtraArguments:    nil,
+		HasMetricsPort:    false,
+	}
+	// DefaultLivenessProbeDaemonSet is definition of the default livenessprobe sidecar for DaemonSet.
+	DefaultLivenessProbeDaemonSet = generator.SidecarConfig{
```
jsafrane, line 122: I think `DefaultHostNetworkLivenessProbe` would be a better name - it's not about Deployment/DaemonSet, but about using host network or not. We had control plane pods using host network in the past and I think we had also daemonset using the pod network.

The same would apply to the sidecar patch files - `assets/common/sidecars/livenessprobe_hostnetwork.yaml`

pkg/driver/common/generator/base_assets.go
```diff
@@ -111,9 +112,15 @@ var (
 			"sidecar.yaml", "common/hypershift/sidecar_add_kubeconfig.yaml.patch",
 		),
 	}
-	// DefaultLivenessProbe is definition of the default livenessprobe sidecar.
-	DefaultLivenessProbe = generator.SidecarConfig{
-		TemplateAssetName: LivenessProbeAssetName,
+	// DefaultLivenessProbeDeployment is definition of the default livenessprobe sidecar for Deployment.
+	DefaultLivenessProbeDeployment = generator.SidecarConfig{
```
jsafrane, line 116: Just `DefaultLivenessProbe`? Or `DefaultPodNetworkLivenessProbe`?

mpatlasov: Makes sense, thank you! I prefer the latter (`DefaultPodNetworkLivenessProbe`) to match host-network name (`DefaultHostNetworkLivenessProbe`).

mpatlasov:
Last force-push only changed names (e.g. DefaultPodNetworkLivenessProbe in place of DefaultLivenessProbeDeployment). No logic changed.

PR #369, opened 2025-03-20
OCPBUGS-54447: Enable required driver services

dkokkino:
The Manila CSI driver, like Cinder, uses a single binary for both the controller and node (worker) services. This can result in services running on unintended hosts. A similar issue was addressed in Cinder in version 4.18, and we should now apply the same fix to Manila.

### DIFF
--- a/assets/overlays/openstack-manila/generated/hypershift/controller.yaml
+++ b/assets/overlays/openstack-manila/generated/hypershift/controller.yaml
@@ -89,6 +89,8 @@ spec:
             weight: 100
       containers:
       - args:
+        - --provide-controller-service=true
+        - --provide-node-service=false
         - --v=${LOG_LEVEL}
         - --cluster-id=${CLUSTER_ID}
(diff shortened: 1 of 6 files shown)


stephenfin:
/lgtm
/hold

hold for CI

mandre (commented):
The bot requires either a related bug in the OCPBUGS project, or someone adding the `qe-approved` label.

> Not mergeable. Needs jira/valid-bug label OR Needs qe-approved label.

This looks like something we want to backport to 4.18, where we started using CSI operator for openstack CSI drivers, so most likely it would be better to move the OSASINFRA-3748 jira to the OCPBUGS project (to allow backporting if we need to).

stephenfin:
/cherry-pick release-4.18

This fixes merge conflicts for backports of https://github.com/openshift/csi-operator/pull/379

stephenfin:
Oh, https://github.com/openshift/csi-operator/pull/375 already exists.

PR #376, opened 2025-04-09
OCPBUGS-54382: Correct ASH driver inject env config

Phaow:
### Correct ASH driver inject env config
- We made a mistake when migrate the azure disk driver operator from `azure-disk-csi-driver-operator` to `csi-operator`, the  `injectEnvAndMounts` which used for ASH the `endpoints` Subpath expected read the `endpoints` value from the cloud configmap key `endpoints` while we use the incorrect volume name for injecting.
 ref -> https://github.com/openshift/azure-disk-csi-driver-operator/blob/release-4.15/pkg/azurestackhub/azure_stack_hub.go#L48-L76
```console
...
2025-03-27T21:07:32.105160383Z W0327 21:07:32.105102       1 azure_disk_utils.go:236] InitializeCloudFromConfig failed with error: read /etc/azure/azurestackcloud.json: is a directory 
...
```

### DIFF
--- a/pkg/driver/azure-disk/azure_disk.go
+++ b/pkg/driver/azure-disk/azure_disk.go
@@ -54,7 +54,7 @@ const (
 	configEnvName         = "AZURE_ENVIRONMENT_FILEPATH"
 	azureStackCloudConfig = "/etc/azure/azurestackcloud.json"
 	// name of volume that contains cloud-config in actual deployment or daemonset
-	podAzureCfgVolumeName = "cloud-config"
+	podAzureCfgVolumeName = "src-cloud-config"
 
 	diskEncryptionSetID = "diskEncryptionSetID"
 


Phaow:
Hi @gnufied @dobsonj , could you help review when you get a chance? Thank you!
/assign @gnufied @dobsonj

gnufied:
I am not sure if this is correct. See, the way `initContainers` in azure-disk csi controller pod, writes merged cloud-config to `cloud-config` emptydir. And that is what we are trying to mount in controller pod. 

I think I changed the code when we merged https://github.com/openshift/azure-disk-csi-driver-operator/blob/release-4.15/pkg/azurestackhub/azure_stack_hub.go#L48-L76 code into `csi-operator` to be uniform between various deployments.

Phaow:
Hi @gnufied , thank you for looking into this! 

> See, the way initContainers in azure-disk csi controller pod, writes merged cloud-config to cloud-config emptydir. And that is what we are trying to mount in controller pod.

Yeah, the initContainers merged config is working as expected, that megre config should be only used for `AZURE_CREDENTIAL_FILE`, while the issue is `AZURE_ENVIRONMENT_FILEPATH` incorrect. It does not broken between 4.16-4.18 due to the `cloud conf` contains `resourceManagerEndpoint: XXXX` , so the azure auth does not goes to read the `endpoints` from our env var `AZURE_ENVIRONMENT_FILEPATH`  -> https://github.com/openshift/azure-disk-csi-driver/blob/release-4.18/vendor/sigs.k8s.io/cloud-provider-azure/pkg/azclient/cloud.go#L130-L136 , while in 4.19 the new driver uses the new sdk and cloud-provider-azure dependency even if the `resourceManagerEndpoint`  exist it always read the customer endpoints form `AZURE_ENVIRONMENT_FILEPATH`  towards file if the env is set -> https://github.com/openshift/azure-disk-csi-driver/blob/master/vendor/sigs.k8s.io/cloud-provider-azure/pkg/azclient/cloud.go#L133-L148C8. From my understanding, it sounds a bug in cloud-provider-azure previously that means if the cloud.conf contains `resourceManagerEndpoint` , then the AZURE_ENVIRONMENT_FILEPATH  overwrite ep does not take effect, it seems not as expected.

gnufied:
hmm, it indeed seems like injector doesn't seem to be copying `endpoints` field and hence it must be copied from original config-map.

/lgtm
/approve

Phaow:
/retitle OCPBUGS-55052: Correct ASH driver inject env config

PR #379, opened 2025-04-15
OCPBUGS-54975: Fix Cinder, Manila driver metrics

stephenfin:
Modify the Cinder operator so that it starts reporting controller driver metrics correctly.

Modify the Manila operator so that we stop creating a controller driver metrics proxy container, since the controller driver does not currently support reporting metrics.


stephenfin:
/retest-required

I haven't been able to reproduce the Cinder failures locally.

PR #380, opened 2025-04-15
NO-JIRA: Normalize generation of driver metrics RBAC proxy sidecar

stephenfin:
The generator has the ability to generate kube-rbac-proxy sidecar containers when exposing driver metrics on the controller or node.
In https://github.com/openshift/csi-operator/pull/379, we noted that while the sidecar containers templates used template variables, the actual driver template used hardcoded strings. This allows for the possibility of bugs if the port numbers are changed at any point.
Resolve this by modifying the generator such that the metric port for the drivers is now templated also. Changes are implemented in multiple steps: I'd encourage reviewers to look at the individual commits.

This includes the commits from https://github.com/openshift/csi-operator/pull/379 and that can as such be considered a dependency.

Dependencies:

- [ ] https://github.com/openshift/csi-operator/pull/379

### DIFF
--- a/assets/common/sidecars/controller_driver_kube_rbac_proxy.yaml
+++ b/assets/common/sidecars/controller_driver_kube_rbac_proxy.yaml
@@ -19,7 +19,7 @@ spec:
           imagePullPolicy: IfNotPresent
           ports:
           - containerPort: ${EXPOSED_METRICS_PORT}
-            name: ${PORT_NAME}
+            name: driver-m
             protocol: TCP
           resources:
             requests:
--- a/assets/common/sidecars/controller_driver_kube_rbac_proxy.yaml
+++ b/assets/common/sidecars/controller_driver_kube_rbac_proxy.yaml
@@ -6,7 +6,7 @@ spec:
   template:
     spec:
       containers:
-        - name: kube-rbac-proxy-${LOCAL_METRICS_PORT}
+        - name: driver-kube-rbac-proxy
           args:
           - --secure-listen-address=0.0.0.0:${EXPOSED_METRICS_PORT}
           - --upstream=http://127.0.0.1:${LOCAL_METRICS_PORT}/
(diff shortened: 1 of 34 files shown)


pkg/generator/asset_generator.go
```diff
@@ -170,15 +166,15 @@ func (gen *AssetGenerator) generateDeployment() error {
...
 		return nil
 	}
 
 	extraReplacements := []string{
-		"${EXPOSED_METRICS_PORT}", strconv.Itoa(int(metricsPort.ExposedPort)),
-		"${LOCAL_METRICS_PORT}", strconv.Itoa(int(metricsPort.LocalPort)),
-		"${PORT_NAME}", metricsPort.Name,
+		"${LOCAL_METRICS_PORT}", strconv.Itoa(int(localMetricsPort)),
+		"${EXPOSED_METRICS_PORT}", strconv.Itoa(int(exposedMetricsPort)),
+		"${PORT_NAME}", "driver-m",
```
mandre, line 182: Is the `${PORT_NAME}` replacement still being used?

assets/common/sidecars/controller_driver_kube_rbac_proxy.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   template:
     spec:
       containers:
-        - name: kube-rbac-proxy-${LOCAL_METRICS_PORT}
+        - name: driver-kube-rbac-proxy
```
mandre, line 9: Have you checked this doesn't break upgrades?

mandre: Sorry for the noise, it's a container name and likely not relevant for upgrade's concerns.

stephenfin:
/retest-required

There's no obvious reason for those clusters not to have come up

stephenfin:
I've dropped the last commit since the rest of these changes are no-ops, as seen in the lack of changes to the assets. I'll propose that last commit separately.

jsafrane:
/lgtm
/verified by CI

There is no difference in the output of the generator, hence there is not much to test.

stephenfin:
/retest-required

Oof

PR #345, opened 2024-12-10
OSASINFRA-3652: openstack-cinder: Use new --with-topology flag

stephenfin:
Rather than relying on the soon-to-be-removed feature gate.

/hold

~This requires https://github.com/kubernetes/cloud-provider-openstack/pull/2743~

*EDIT:* This is now available on the `release-4.18` and `release-4.19` branches https://github.com/openshift/cloud-provider-openstack/commit/6cb478d03f86ebce6e36b8ea402a3147cddb3eec

PR #383, opened 2025-05-13
OCPBUGS-56128: openstack-cinder: Set --with-topology flag for node driver also

stephenfin:
This is a follow-up to https://github.com/openshift/csi-operator/pull/345. Failure to do so results in the node driver reporting topology information which the provisioner respects, preventing us attaching the created volume.


stephenfin:
Failures were because I forgot to run `make update`. That's what I get for trying to fire something off after hours... :roll_eyes: Should be resolved now.

PR #389, opened 2025-06-05
STOR-2479: Add CredentialsRequest for the EFS node DaemonSet

jsafrane:
AWS EFS CSI driver will need a new identity + role in AWS for the node DaemonSet to support zonal volumes.

* Extend the operator to be able to handle CredentialsRequest in the guest assets.
* Add CredentialsRequest to the EFS operator + make sure the CSI driver gets it in the right format (it should be copy+paste of the controller initContainer)
  * When user / OCP console sets just ROLEARN (with the controller ARN), do not  create the node CredentialsRequest and don't expect the node Secret exists. **This allow users to upgrade with just regional volumes support** and then can add `NODE_ROLEARN` env. var to their Subscription later.
    * This needs https://github.com/openshift/library-go/pull/1974 merged.
* Use `dnsPolicy: Default` to turn off in-cluster DNS. The driver's DNS queries would be resolved by  a random DNS pod in a random zone, and sometimes resolution of a local volume would succeed (when the DNS pod is in the same zone as the volume) and sometimes it would fail.

Until now, the DaemonSet used node's AWS identity, but that one does not have `elasticfilesystem:DescribeMountTargets` and it's not possible to add it there during cluster upgrade. And it's better to use our own fine-grained role instead of a bigger common one anyway.

* Update library-go to get https://github.com/openshift/library-go/pull/1974, bump go and Kubernetes libs as result.


jsafrane:
/retest-required
This PR should not change azure...

jsafrane:
/hold
for https://github.com/openshift/library-go/pull/1974 to merge first

jsafrane:
/hold cancel
I bumped library-go with the fix.

jsafrane:
rebased

PR #401, opened 2025-06-23
Draft: AWS standalone cross account doc

RomanBednar:
>Do not merge, this PR is for collaboration purposes only and will be closed later.

## Notes  
  
- Documentation  
  - Sections to remove from our documentation:  
        - `5.9.5. AWS EFS CSI cross account support`  
        - `5.9.6. Creating and configuring access to EFS volumes in AWS`  
        - These two sections should be consolidated into a single comprehensive section, as they are interdependent and cannot function independently  
  - A statement should be added indicating that this functionality is fully supported for both STS and non-STS clusters  
  - Code blocks with comments should be replaced with asides/alerts/callouts for improved readability  
  - Section `5.9.8. Creating static PVs with Amazon Elastic File Storage` should mention that when using Cross Account feature the volumeHandle has to reference EFS ID of the filesystem in Account B  
  
- Manual testing procedures:  
  - Test standalone cluster only (no ROSA) with STS and without STS enabled, the docs "branch" to steps:  
      - 5.9.5.5.1 - for non-STS clusters  
      - 5.9.5.5.2 - for STS clusters  
  - Follow the guide precisely, utilizing copy-paste methodology where applicable  
  - Verify successful execution of each command before proceeding to the next step  
  - Document any unclear instructions for future refinement  
  
- CI configuration considerations:  
  - Ideally, CI coverage should align with our documentation; however, current implementation presents some discrepancies:  
      - EFS Operator installations in CI are not handled via ccoctl due to shared CI steps with ROSA → addressing this would require creating duplicate CI steps for standalone environments or avoiding ccoctl for standalone configurations  
      - Certain CI components are optimized for performance and may not precisely reflect every documented command. For example, subnet names can be loaded as known values from previous steps rather than executing repeated AWS CLI calls: https://github.com/openshift/release/pull/65233/files#diff-3eef242bf0070d5c2dc96f9d2037742f250347ed56c2faa6363ab1ed74a64c1eR90

## Testing

- Install EFS operator from OperatorHub as usual, for STS use ccoctl to create AWS Role

- Create testing namespace:
```
oc create ns testns
```

- Create PVC:
```
cat << EOF | oc apply -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: efs-crossaccount-pvc
  namespace: testns
spec:
  accessModes:
    - ReadWriteMany
  storageClassName: efs-sc-cross
  resources:
    requests:
      storage: 1Gi
EOF
```

- Create pod to mount the volume:
```
cat << EOF | oc apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: example-1
  labels:
    app: httpd
  namespace: testns
spec:
  volumes:
    - name: task-pv-storage
      persistentVolumeClaim:
        claimName: efs-crossaccount-pvc
  securityContext:
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault
  containers:
    - name: httpd
      image: 'image-registry.openshift-image-registry.svc:5000/openshift/httpd:latest'
      ports:
        - containerPort: 8080
      volumeMounts:
        - mountPath: "/usr/share/nginx/html"
          name: task-pv-storage
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
            - ALL
EOF
```

- Check if there are no events/errors and that volume mounted

---

```
cat << EOF | oc create -f -
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: efs-sc
provisioner: efs.csi.aws.com
parameters:
  provisioningMode: efs-ap 
  fileSystemId: fs-0b55afffece6c2d6c 
  directoryPerms: "700" 
  gidRangeStart: "1000" 
  gidRangeEnd: "2000" 
  basePath: "/dynamic_provisioning"
  volumeBindingMode: Immediate
EOF
```

### DIFF
--- a/README-tmp.md
+++ b/README-tmp.md
@@ -43,14 +43,29 @@ mkdir -p $SCRATCH_DIR
 oc whoami
 ```
 
-4. Configure AWS CLI profiles as environment variables for account switching:
+4. Determine OpenShift cluster flavor and set node selector
+
+EFS Cross Account feature requires assigning AWS IAM policies to nodes running EFS CSI controller pods, however this is
+not consistent for every OpenShift flavor.
+
+If your cluster is deployed as Hosted Control Plane (HyperShift), set NODE_SELECTOR environment variable to hold worker node label:
+```bash
+export NODE_SELECTOR=node-role.kubernetes.io/worker
+```
+
+For all other OpenShift flavors set NODE_SELECTOR environment variable to hold master node label:
+```bash
+export NODE_SELECTOR=node-role.kubernetes.io/master
+```
+
+5. Configure AWS CLI profiles as environment variables for account switching:
 
 ```bash
 export AWS_ACCOUNT_A="<ACCOUNT_A_NAME>"
 export AWS_ACCOUNT_B="<ACCOUNT_B_NAME>"
 ```
 
-5. Ensure that your AWS CLI is configured with JSON output format as the default for both accounts:
+6. Ensure that your AWS CLI is configured with JSON output format as the default for both accounts:
 
 ```bash
 export AWS_DEFAULT_PROFILE=${AWS_ACCOUNT_A}
@@ -61,7 +76,7 @@ aws configure get output
 
 If the AWS CLI commands above return no value, the default output format is already set to JSON and no changes are required. If any other value is returned, reconfigure your AWS CLI to use JSON format. For detailed instructions on changing output formats, refer to: https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html
 
-6. Unset `AWS_PROFILE` in your shell to prevent conflicts with `AWS_DEFAULT_PROFILE`:
+7. Unset `AWS_PROFILE` in your shell to prevent conflicts with `AWS_DEFAULT_PROFILE`:
 
 ```bash
 unset AWS_PROFILE
@@ -193,7 +208,7 @@ EOF
 
 ```bash
 EFS_POLICY_ARN=arn:aws:iam::aws:policy/AmazonElasticFileSystemClientFullAccess
-for NODE in $(oc get nodes --selector=node-role.kubernetes.io/master | tail -n +2 | awk '{print $1}')
+for NODE in $(oc get nodes --selector=${NODE_SELECTOR} | tail -n +2 | awk '{print $1}')
 do
    INSTANCE_PROFILE=$(aws ec2 describe-instances --filters "Name=private-dns-name,Values=${NODE}" --query 'Reservations[].Instances[].IamInstanceProfile.Arn' --output text | awk -F'/' '{print $NF}')
    MASTER_ROLE_ARN=$(aws iam get-instance-profile --instance-profile-name ${INSTANCE_PROFILE}  --query 'InstanceProfile.Roles[0].Arn' --output text)


README-tmp.md
```diff
@@ -0,0 +1,379 @@
...
+```
+
+4. Configure AWS CLI profiles as environment variables for account switching:
+
+```bash
+export AWS_ACCOUNT_A="<ACCOUNT_A_NAME>"
+export AWS_ACCOUNT_B="<ACCOUNT_B_NAME>"
+```
+
+5. Ensure that your AWS CLI is configured with JSON output format as the default for both accounts:
```
Phaow, line 52:
```suggestion
5. Ensure that your AWS CLI is configured with JSON output format as the default for both accounts(the output format could be override by `export AWS_DEFAULT_OUTPUT="json"`):
```
Maybe better.

README-tmp.md
```diff
@@ -0,0 +1,379 @@
...
+```bash
+export AWS_DEFAULT_PROFILE=${AWS_ACCOUNT_B}
+for ROUTE_TABLE_ID in $(aws ec2 describe-route-tables   --filters "Name=vpc-id,Values=${AWS_ACCOUNT_B_VPC_ID}"   --query "RouteTables[].RouteTableId" | jq -r '.[]')
+do
+    echo Route table ID is $ROUTE_TABLE_ID
+    aws ec2 create-route --route-table-id ${ROUTE_TABLE_ID} --destination-cidr-block ${AWS_ACCOUNT_A_VPC_CIDR} --vpc-peering-connection-id ${PEER_REQUEST_ID}
+done
+```
+
+5. Enable DNS resolution for Account A to access Account B's VPC:
```
Phaow, line 285: I do not add the step(optional step?) in CI, from the [upstream doc](https://github.com/openshift/aws-efs-csi-driver/tree/master/examples/kubernetes/cross_account_mount) it seems also needs if we want to ensure that your EFS Mount Target is in the same availability zone with node, it also needs extra config in secret right?

> If you would like to ensure that your EFS Mount Target is in the same availability zone as your EKS Node, then ensure you have completed the [prerequisites for cross-account DNS resolution](https://github.com/aws/efs-utils?tab=readme-ov-file#crossaccount-option-prerequisites) and include the crossaccount key with value true. For example, kubectl create secret generic x-account --namespace=kube-system --from-literal=awsRoleArn='arn:aws:iam::123456789012:role/EFSCrossAccountAccessRole' --from-literal=crossaccount='true' instead.

README-tmp.md
```diff
@@ -0,0 +1,379 @@
...
+--region ${AWS_REGION} \
+--encrypted | jq -r '.FileSystemId') \
+&& echo $CROSS_ACCOUNT_FS_ID
+```
+
+3. Configure region-wide Mount Targets for EFS (this creates a mount point in each subnet of your VPC):
+
+```bash
+for SUBNET in $(aws ec2 describe-subnets \
+  --query 'Subnets[*].{SubnetId:SubnetId}' \
```
Phaow, line 331:
```suggestion
  --filters "Name=vpc-id,Values=${AWS_ACCOUNT_B_VPC_ID}" \
  --query 'Subnets[*].{SubnetId:SubnetId}' \
```
We'd better filter with vpc id since current way will query all subnets in the region.

README-tmp.md
```diff
@@ -0,0 +1,379 @@
...
+
+##  5.9.5.9 Configure EFS operator for cross-account access
+
+1. Define custom names for the secret and storage class to be created in subsequent steps:
+```bash
+export SECRET_NAME=my-efs-cross-account
+export STORAGE_CLASS_NAME=efs-sc-cross
+```
+
+2. Create a secret that references the role ARN in Account B:
```
Phaow, line 350: Maybe we'd better also mention that `openshift-cluster-csi-drivers` is the operator installed namespace, sometimes customer install it in some other namespace.

README-tmp.md
```diff
@@ -0,0 +1,379 @@
+# 5.9.5 Configuring EFS Cross Account
+
+## 5.9.5.1 Requirements
+
+- The EFS CSI Operator must be installed prior to beginning this procedure. Follow the _Installing the AWS EFS CSI Driver Operator_ procedure.
+- Both the Red Hat OpenShift Service on AWS cluster and EFS file system must be located in the same AWS region.
+- Ensure that the two VPCs referenced in this guide utilize different network CIDR ranges.
```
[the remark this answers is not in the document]
RomanBednar: I don't think so, similar to the "same availability zone" feature mentioned in another comment here - we don't document it now and probably should not add new features in scope of this change.

README-tmp.md
RomanBednar: Yes this is one of the methods of changing the output format, however I intentionally avoided explaining it here since we would be documenting AWS here, which is not desired. For this purposes I've added a link in the paragraph below to official AWS doc (which also explains AWS_DEFAULT_OUTPUT var): https://github.com/openshift/csi-operator/pull/401/files#diff-5c4f1f33bb4c41d6c3e57b67e95dc097ebf581e3fdc2f5bbf848f15135179035R61

README-tmp.md
RomanBednar: Good point, it seems like this would be a prerequisite to using `crossaccount='true'` which is indeed reflected in a secret -> but this is something we do not document currently. So I would remove this step for now unless we get an RFE just so we don't introduce this change and bloat the testing even more.

RomanBednar:
@Phaow Thank you for the comments and review, I've addressed them now - please take a look and let me know if there's anything else. When done I suggest we test this manually to make sure there are no mistakes before handoff to docs team.

README-tmp.md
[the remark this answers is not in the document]
Phaow: Acked and thanks for the double confirm!

Phaow:
LGTM, @ropatil010 as the feature main QE owner, could you please cross check and verify the doc works as expected? Thanks!

README-tmp.md
```diff
@@ -0,0 +1,374 @@
...
+      "Effect": "Allow",
+      "Action": "sts:AssumeRole",
+      "Resource": "${ACCOUNT_B_ROLE_ARN}"
+    }
+  ]
+}
+EOF
+```
+
+3. In AWS Account A, attach the AWS-managed policy "AmazonElasticFileSystemClientFullAccess" to the OpenShift Container Platform cluster master role:
```
ropatil010, line 192: Currently tested on OCP standalone and OCP HCP kind clusters.
Note: If running on HCP kind cluster no need of executing this 3rd steps. Since HCP cluster has only worker nodes. 
I have tested with all mentioned steps on HCP kind cluster except this 3rd step and we can provision volume, able to read/write/execute. 
We can mark this doc on both OCP standalone and OCP HCP kind clusters. 
cc: @jsafrane @Phaow

Phaow: So do you test without the step volumeReclaimPolicy work as expected? I mean in hcp clusters delete the pvc then the pv should be deleted to be consistent with the `volumeReclaimPolicy:Delete`. The step is used for driver controller reclaim volumes.

Phaow: Yes, it is up to which nodes the driver controller pod located(hcp clusters the driver controller located at workers), thanks for the doulbe checking!

ropatil010:
Added test results here: https://issues.redhat.com/browse/STOR-2378 
Based on the testing adding label.  
/lgtm

PR #390, opened 2025-06-05
STOR-2477: create-efs-volume: add single-zone support

jsafrane:
`create-efs-volume start --single-zone=us-east-1a` creates a single-zone AWS EFS volume in given zone.

PR #404, opened 2025-07-09
STOR-2477: Correct efs single-zone storageclass

Phaow:
- We introduce the `create-efs-volume` single zone support in https://github.com/openshift/csi-operator/pull/390, while when storageclass set with `allowedTopologies` must using `volumeBindingMode: WaitForFirstConsumer`(**[the default](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/storage/v1/defaults.go#L39)** is `Immediate`) otherwise the volume provision will not respect the allowedTopologies settings.

### DIFF
--- a/assets/overlays/aws-efs/testing/sc-single-zone.yaml
+++ b/assets/overlays/aws-efs/testing/sc-single-zone.yaml
@@ -15,3 +15,4 @@ allowedTopologies:
   - key: topology.kubernetes.io/zone
     values:
     - ${zone}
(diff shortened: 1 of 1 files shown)


Phaow:
Hi @jsafrane , could you please help take a look when you get a chance? Thank you!
/assign @jsafrane

Phaow:
The PR is just using in CI test configuration no user facing changes, we're safe to merge.
/label acknowledge-critical-fixes-only

PR #432, opened 2025-09-10
OCPBUGS-54385: Avoid informer resync

mpatlasov:
https://issues.redhat.com/browse/OCPBUGS-54385

Most of controllers do per-controller resync, so informer resync is redundant. The only exception is LogLevel controller, but it must be safe to call reconciliation loop when an update happens without resync.


duanwei33:
The pre-merge verification for the open PR passed. It shows a rate of 2 records per minute, with a total of 128 records generated in one hour, and no additional records every 10 minutes:
```
$ oc -n openshift-cluster-csi-drivers logs smb-csi-driver-operator-6478c6dfff-9s65x --since=1h  | grep "reconciliation failed" | wc -l
128
```

PR #434, opened 2025-09-24
NO-JIRA: Normalize generation of driver metrics RBAC proxy sidecar (extended)

stephenfin:
This builds on #380 by changing the name of the proxy sidecar used for driver metrics.


stephenfin:
I've just rebased this on top of master, to catch any changes that might affect this. This is otherwise identical to previously. Hopefully CI is happier now :crossed_fingers:

PR #435, opened 2025-09-25
STOR-2627: Enable nodeAllocatableUpdatePeriodSeconds in AWS EBS

jsafrane:
AWS EBS CSI driver should support `MutableCSINodeAllocatableCount` feature.

Set CSIDriver `nodeAllocatableUpdatePeriodSeconds` to 10 minutes when the feature gate is enabled. Kubelet will then call NodeGetDriverInfo every 10 minutes to update the attach limit of EBS volumes.

cc @openshift/storage


pkg/driver/aws-ebs/aws_ebs.go
```diff
@@ -208,6 +211,14 @@ func GetAWSEBSOperatorControllerConfig(ctx context.Context, flavour generator.Cl
 		cfg.ExtraControlPlaneControllers = append(cfg.ExtraControlPlaneControllers, ctrl)
 	}
 
+	cfg.ExtraReplacementsFunc = func() []string {
```
Phaow: Hi @jsafrane , I did some tests for enable nodeAllocatableUpdatePeriodSeconds in ebs csi driver and found that we still needs https://github.com/openshift/csi-external-attacher/pull/89 which contains(https://github.com/kubernetes-csi/external-attacher/commit/462cd5450e29f746fef620a6ab53b9f67c2c16c6 we needed for MutableCSINodeAllocatableCount featuregate args in external attacher). In addition it seems the `cfg.ExtraReplacementsFunc` only do the replace the first time in cluster install when I enable the TechPreviewNoUpgrade it does not update the csidriver, I needs to delete the `csidriver` trigger the resource reconcile and update it with `nodeAllocatableUpdatePeriodSeconds: 600`. This should be an issue when upgrade/enable featuregate. Could you help check when you get a chance? Thank you! ^^

```console
# After enable the TechPreviewNoUpgrade the csidriver does not update
$ oc get csidriver ebs.csi.aws.com -oyaml
apiVersion: storage.k8s.io/v1
kind: CSIDriver
metadata:
  annotations:
    csi.openshift.io/managed: "true"
    operator.openshift.io/spec-hash: d37d616b853448c2d019d19e3987ebfb751f3da09de52be009406f4d8f8fd7d6
  creationTimestamp: "2025-10-23T06:09:33Z"
  name: ebs.csi.aws.com
  resourceVersion: "86418"
  uid: 8ac66b43-1f14-4be7-9d4b-1e65f29ed14a
spec:
  attachRequired: true
  fsGroupPolicy: File
  podInfoOnMount: false
  requiresRepublish: false
  seLinuxMount: true
  storageCapacity: false
  volumeLifecycleModes:
  - Persistent

# delete the csidriver trigger reconcile
$ oc delete csidriver ebs.csi.aws.com
csidriver.storage.k8s.io "ebs.csi.aws.com" deleted
$ oc get csidriver ebs.csi.aws.com -oyaml
apiVersion: storage.k8s.io/v1
kind: CSIDriver
metadata:
  annotations:
    csi.openshift.io/managed: "true"
    operator.openshift.io/spec-hash: d37d616b853448c2d019d19e3987ebfb751f3da09de52be009406f4d8f8fd7d6
  creationTimestamp: "2025-10-23T07:01:02Z"
  name: ebs.csi.aws.com
  resourceVersion: "116117"
  uid: 84d59f30-7b5f-4059-9cb5-a02089c2093b
spec:
  attachRequired: true
  fsGroupPolicy: File
  nodeAllocatableUpdatePeriodSeconds: 600
  podInfoOnMount: false
  requiresRepublish: false
  seLinuxMount: true
  storageCapacity: false
  volumeLifecycleModes:
  - Persistent
```

jsafrane: You are correct, the external-attacher rebase is necessary too.

> it seems the cfg.ExtraReplacementsFunc only do the replace the first time in cluster install when I enable the TechPreviewNoUpgrade it does not update the csidriver

I'll look into this. The operator _should_ restart when something changes FeatureGates...

PR #552, opened 2026-05-12
OCPBUGS-85457: Add init container for Manila node daemonset

mandre:
The Manila CSI driver node plugin fatally exits on startup when the NFS CSI plugin socket (`/var/lib/kubelet/plugins/csi-nfsplugin/csi.sock`) is not yet available. After node disruptions, both DaemonSets restart concurrently and the Manila driver crashes because upstream ProbeForever only retries on DeadlineExceeded, not on Unavailable (connection refused). This causes excessive container restarts (>3), failing the CI invariant test.

In addition to an upcoming fix to make the driver more tolerant, add a `wait-for-nfs-plugin` init container to the Manila node DaemonSet that polls for the NFS plugin socket before allowing the csi-driver container to start. This ensures proper startup ordering between the two DaemonSets without requiring changes to the upstream driver.

### DIFF
--- a/assets/overlays/openstack-manila/patches/node_add_driver.yaml
+++ b/assets/overlays/openstack-manila/patches/node_add_driver.yaml
@@ -90,6 +90,28 @@ spec:
               cpu: 10m
               memory: 50Mi
           terminationMessagePolicy: FallbackToLogsOnError
+      initContainers:
+        - name: wait-for-nfs-plugin
+          image: ${DRIVER_IMAGE}
+          command:
+            - sh
+            - -c
+            - |
+              until [ -S /var/lib/kubelet/plugins/csi-nfsplugin/csi.sock ]; do
+                echo "Waiting for NFS CSI plugin socket..."
+                sleep 2
+              done
+          securityContext:
+            readOnlyRootFilesystem: true
+            privileged: false
+          resources:
+            requests:
+              cpu: 10m
+              memory: 50Mi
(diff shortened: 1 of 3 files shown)


mandre:
Infra issue...
/test e2e-openstack-manila-csi

mandre:
The hypershift jobs are currently broken
/override ci/prow/hypershift-e2e-openstack-aws-csi-manila

assets/overlays/openstack-manila/patches/node_add_driver.yaml
```diff
@@ -90,6 +90,28 @@ spec:
               cpu: 10m
               memory: 50Mi
           terminationMessagePolicy: FallbackToLogsOnError
+      initContainers:
```
stephenfin, line 93: Any chance we could file this as an upstream bug and add a link to that here? We'll want to backport this so this approach makes sense for now, but I believe it would be a good idea to remove this in the future if possible.

stephenfin: Discussed this offline and this isn't an either-or thing: we actually want both fixes. This change to add an explicit order for containers to start is correct (and not expensive) when we know one depends on the other. Let's capture that instead (via a comment or the commit message) for future us

mandre: I've reported the issue for the manila driver at https://github.com/kubernetes/cloud-provider-openstack/issues/3111.

PR #573, opened 2026-07-03
STOR-3030: feat(operator): detect GCP Dedicated and use hyperdisk-balanced StorageClass

jsafrane:
At operator startup, fetch the Infrastructure CR (with retry for up to 1 minute during early cluster bootstrap) and check if the GCP region starts with "u-". If so, the cluster runs on GCP Dedicated where only hyperdisk-balanced disks are supported — create only the hyperdisk-balanced StorageClass as default.

On regular GCP, the existing standard-csi and ssd-csi StorageClasses are used as before.

### DIFF
--- a/legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
+++ b/legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
@@ -12,6 +12,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	kubeclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -60,6 +61,10 @@ const (
 	ocpDefaultLabelFmt = "kubernetes-io-cluster-%s=owned"
 
 	operatorImageVersionEnvVarName = "OPERATOR_IMAGE_VERSION"
+
+	// gcpDedicatedRegionPrefix is the prefix for GCP Dedicated regions.
+	// GCP Dedicated regions start with "u-" (e.g. "u-germany-northeast1").
+	gcpDedicatedRegionPrefix = "u-"
 )
 
 func RunOperator(ctx context.Context, controllerConfig *controllercmd.ControllerContext) error {
@@ -243,22 +248,22 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		dynamicClient,
 		assets.ReadFile,
 		"servicemonitor.yaml",
-	).WithStorageClassController(
+	)
+
+	storageClassFiles, err := getStorageClassFiles(ctx, configClient)
+	if err != nil {
+		return err
+	}
+
+	csiControllerSet = csiControllerSet.WithStorageClassController(
 		"GCPPDDriverStorageClassController",
 		assets.ReadFile,
-		[]string{
-			"storageclass.yaml",
-			"storageclass_ssd.yaml",
-		},
+		storageClassFiles,
 		kubeClient,
 		kubeInformersForNamespaces.InformersFor(""),
 		operatorInformers,
 		getKMSKeyHook(operatorInformers.Operator().V1().ClusterCSIDrivers().Lister()),
 	)
-
-	if err != nil {
-		return err
-	}
 
 	klog.Info("Starting the informers")
 	go kubeInformersForNamespaces.Start(ctx.Done())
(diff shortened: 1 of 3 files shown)


legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
```diff
@@ -274,6 +279,49 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+// early cluster installation the CR may not exist yet.
+// On GCP Dedicated, only hyperdisk-balanced is supported.
+// On regular GCP, standard-csi and ssd-csi are used.
+func getStorageClassFiles(ctx context.Context, configClient configclient.Interface) ([]string, error) {
+	regularFiles := []string{
+		"storageclass.yaml",
+		"storageclass_ssd.yaml",
+	}
+	gcpDedicatedFiles := []string{
+		"storageclass_hyperdisk_balanced.yaml",
```
mpatlasov, line 294: You forgot to cherry-pick https://github.com/openshift/gcp-pd-csi-driver-operator/pull/196/changes/5e82dc18552eb30e42b86fe3978f3b51c4576f08

legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
```diff
@@ -243,23 +248,23 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+	}
+
+	csiControllerSet = csiControllerSet.WithStorageClassController(
 		"GCPPDDriverStorageClassController",
 		assets.ReadFile,
-		[]string{
-			"storageclass.yaml",
-			"storageclass_ssd.yaml",
-		},
+		storageClassFiles,
```
mpatlasov, line 261: I think we also need to **delete** older StorageClasses (`standard-csi` and `ssd-csi`) if we detected "Dedicated" case.

Otherwise, if a customer installs any previous version of OCP in "Dedicated" environment, and then upgrades to this version, all three StorageClasses would exist: `standard-csi` and `ssd-csi` and `hyperdisk-balanced`. Also, `standard-csi` would keep "default: true" annotation, and `hyperdisk-balanced` loose it:

```
$ oc get sc hyperdisk-balanced -o yaml
allowVolumeExpansion: true
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "false"
```

legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
```diff
@@ -274,6 +279,49 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+		"storageclass.yaml",
+		"storageclass_ssd.yaml",
+	}
+	gcpDedicatedFiles := []string{
+		"storageclass_hyperdisk_balanced.yaml",
+	}
+
+	var region string
+	var lastErr error
+	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
```
mpatlasov, line 299: Hard-coded `1*time.Minute` looks weird. It can work in our tests, but what if in customer env one minute is not enough. I think the situation when we **never** can Get() Infrastructure object means the cluster is ultimately broken and it doesn't matter if the operator sits in the loop or bails out with error. Hence, it must be safer to iterate over `Infrastructures().Get()` forever (until it succeeds). Can you see any scenario when returning error after one minute loop is better than looping forever?

legacy/gcp-pd-csi-driver-operator/pkg/operator/storageclass_test.go
```diff
@@ -0,0 +1,129 @@
...
+	configClient := fakeconfig.NewSimpleClientset()
+	// Use a short timeout so the retry loop finishes quickly in tests.
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	_, err := getStorageClassFiles(ctx, configClient)
+	if err == nil {
+		t.Fatal("expected error when Infrastructure CR does not exist, got nil")
+	}
+}
```
mpatlasov, line 129: It's not nice that CI skips these tests. I opened https://github.com/openshift/csi-operator/pull/574 fixing this issue.

legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
jsafrane: It will cause the operator pod crashlooping and thus it is somehow visible to a cluster admin.
If it stays in the loop forever, nobody will notice anything suspicious.

We already do something similar with feature gates:
https://github.com/openshift/csi-operator/blob/723e2b4ba49dea9027085ba78f07463084eb939d/legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go#L104-L111

legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
jsafrane: We did not support GCP dedicated in the previous release.

In addition, if someone installed it there, they must have created some storage classes on their own to make storage work. This patch just adds `hyperdisk-dedicated` to the cluster and marks it as default, if there is no default.

legacy/gcp-pd-csi-driver-operator/pkg/operator/storageclass_test.go
jsafrane: thanks!

legacy/gcp-pd-csi-driver-operator/pkg/operator/starter.go
jsafrane: That's a really good point. There is no GCP dedicated CI (yet) that would catch it.

jsafrane: Fixed

mpatlasov:
@jsafrane , the new file `storageclass_hyperdisk_balanced.yaml` is good, but it is placed wrongly. It needs to be in `legacy/gcp-pd-csi-driver-operator/assets`.

mpatlasov:
/override ci/prow/hypershift-aws-e2e-external
The test fails to create cluster:
```
    hypershift_framework.go:518: failed to create cluster, tearing down: failed to create infra: failed to create VPC: operation error EC2: CreateVpc, https response error StatusCode: 400, RequestID: 9205a86b-e282-4988-99be-ac5c4259a69a, api error VpcLimitExceeded: The maximum number of VPCs has been reached.
```
It is not relevant to GCP PD CSI operator.

PR #237, opened 2024-07-10
STOR-1856: STOR-1606: migrate legacy AWS EFS operator code to csi-operator

RomanBednar:
Followup cleanup PRs:
- https://github.com/openshift/csi-operator/pull/244
- https://github.com/openshift/release/pull/54571
- https://github.com/openshift-eng/ocp-build-data/pull/5120

PR #217, opened 2024-04-19
STOR-1726: auto-generate assets for CIFS/Samba driver

mpatlasov:
The patch ensures that all assets in `overlays/samba/generated/standalone` are generated by `generator`. Resulting assets survive simple dynamic-provisioning test with `test/e2e/samba/samba-server.yaml`

PR #205, opened 2024-04-04
STOR-1726: Minimal implementation of CIFS/Samba CSI driver operator

mpatlasov:
This PR implements the minimal set of go-files and yaml-assets to enable the command
```
./bin/csi-driver-smb-operator start --kubeconfig $KUBECONFIG --namespace openshift-cluster-csi-drivers
```
to install and run CIFS/Samba CSI driver:
```
$ oc get po -n openshift-cluster-csi-drivers
csi-smb-node-5hvqm                           3/3     Running   0          14s
csi-smb-node-f7mn2                           3/3     Running   0          14s
smb-csi-driver-controller-8657649897-dw47d   3/3     Running   0          16s
smb-csi-driver-controller-8657649897-fht45   3/3     Running   0          16s
```

@openshift/storage

PR #576, opened 2026-07-08
STOR-2997: Minimal implementation of GCP PD CSI driver operator

mpatlasov:
This PR implements the minimal set of go-files and yaml-assets to enable the command
```
./bin/gcp-pd-csi-driver-operator start --kubeconfig $KUBECONFIG --namespace openshift-cluster-csi-drivers
```
to install and run GCP PD CSI driver:
```
gcp-pd-csi-driver-controller-c949b748-xkljz   10/10   Running   0          3m2s
gcp-pd-csi-driver-controller-c949b748-xvgl9   10/10   Running   0          3m2s
gcp-pd-csi-driver-node-5kcw5                  3/3     Running   0          3m2s
gcp-pd-csi-driver-node-frl74                  3/3     Running   0          3m2s
gcp-pd-csi-driver-node-gf9k7                  3/3     Running   0          3m2s
gcp-pd-csi-driver-node-p95sv                  3/3     Running   0          3m2s
gcp-pd-csi-driver-node-q4r98                  3/3     Running   0          3m2s
gcp-pd-csi-driver-node-xvmgq                  3/3     Running   0          3m2s
```
/cc @openshift/storage


pkg/driver/gcp-pd/gcp_pd.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/openshift/csi-operator/pkg/operator/config"
+
+	opv1 "github.com/openshift/api/operator/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
+	generatedAssetBase = "overlays/gcp-pd/generated"
+)
+
```
rhrmo, line 60: I see we have the generated assets but there is no `generator.CSIDriverGeneratorConfig` function, these should be generated from this function, right? Or am I missing something?

RomanBednar:
PR seems somewhat incomplete - when I delete `assets/overlays/gcp-pd/generated` directory and try to re-generate the assets it fails:

```
make update-generated-assets
hack/update-generated-assets.sh
assets/assets.go:7:28: pattern overlays/*: cannot embed directory overlays/gcp-pd: contains no embeddable files
make: *** [update-generated-assets] Error 1
```

This is because `collectConfigs()` is not wired to load `CSIDriverGeneratorConfig` (which is missing) - this is what @rhrmo pointed out. 

Also, there's a typo in the binary in PR description: `./bin/gcp-pd-csi-driver-smb-operator` -> smb

I was thinking that this PR could be an intermediate step where you just intend to demonstrate a working binary - however the switch from legacy directory in Containerfile would indicate that the migration to generated files is finished. 

When I worked on AWS migration I implemented this in a single [PR](https://github.com/openshift/csi-operator/pull/237), not saying it's the only right way but I think that first we need to get the generators working and see CI tests are passing with those assets. Then cross-check the assets with the legacy ones - it's easy to misss something.

mpatlasov:
@RomanBednar , @rhrmo , thank you for review, I fixed a typo in PR description -- nice catch!

You are right, the PR is intermediate. It provides the operator binary that works in manual tests and passes CI, but based on pre-defined assets. The next step is to auto-generate assets with generator machinery. SMB operator went through the same process: compare https://github.com/openshift/csi-operator/pull/205 vs. https://github.com/openshift/csi-operator/pull/217.

> however the switch from legacy directory in Containerfile would indicate that the migration to generated files is finished

I'd prefer to make this switch as early as possible to see CI failures (and any other complains) earlier and have time to fix them before the end of the cycle. Let me know if you prefer other approach -- keep the PR open until everything is done and polished before the merge.

pkg/driver/gcp-pd/storageclass_test.go
```diff
@@ -0,0 +1,129 @@
...
+			if len(files) != len(test.expectedFiles) {
+				t.Fatalf("expected %d files, got %d: %v", len(test.expectedFiles), len(files), files)
+			}
+			for i, f := range files {
+				if f != test.expectedFiles[i] {
+					t.Errorf("file[%d]: expected %q, got %q", i, test.expectedFiles[i], f)
+				}
+			}
+		})
+	}
```
rhrmo, line 116: Just a suggestion, could we possibly include [assert library](https://pkg.go.dev/github.com/stretchr/testify/assert) and use it for unit tests? I think we lack use of this library in our repos and for unit tests this is a great tool.

Then the code could look for example something like this (this is just an example, it can definitely be done better):
```
for _, test := range tests {
                t.Run(test.name, func(t *testing.T) {
                        configClient := fakeconfig.NewSimpleClientset(test.infra)
                        ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
                        defer cancel()

                        files, err := getStorageClassFiles(ctx, configClient)

                        //using assert
                        assert.NoErrorf(t, err, "unexpected error: %v", err)
                        assert.Len(t,test.expectedFiles, len(files), "wrong number of files")
                        assert.Equal(t, test.expectedFiles, files, "files don't match expected")
                })
        }
```

rhrmo:
@mpatlasov I don't mind if we don't have the code to generate assets right now, I was just unsure why it was not there yet, we can merge this PR without the generation code. I left a unit test suggestion, but we do not need to switch to the library right now as getting this PR merged is a priority. I don't think the e2e tests are failing because of changes from this pr and required ones seem to be passing. 
/lgtm

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-csi

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-serial

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-serial

RomanBednar:
> Let me know if you prefer other approach

I don't have a strong opinion in that, I just fail to see what's the benefit of the phased approach :) why would we want manually created assets in "generated" directory at any point? It's the same approach as having the assets from GCP PD operator repo copied here to legacy dir and testing with those right? That way you know CI setup and builds are ok, and then I would start generating the assets and comparing them to legacy (static) ones. Or are the legacy files anyhow different from the current ones in "generated" dir? If so, how and why? 

I think we can merge this as long as the generator skips GCP PD (which it does now) so it does not get in way of other PRs and `verify-generated-assets` check. I'm just trying to understand the reasoning.

> I'd prefer to make this switch as early as possible

Yes, it's easy to miss something. So you want to merge this first and then start working on the generator stuff? I think the approach is fine, but it's a significant portion of this effort. I remember it was quite tricky to convince the generator to do the right thing when I migrated EFS a while ago.

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-csi

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-serial

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-csi

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-serial

mpatlasov:
> It's the same approach as having the assets from GCP PD operator repo copied here to legacy dir and testing with those right? 

@RomanBednar ,  you are right, in this PR asset yaml files in `generated` dir are exactly the same as in `legacy/...gcp.../assets`. However, the way how these assets look like runtime, on API-server, depends on custom go code we currently have in `legacy/...gcp.../starter.go`. Re-implementing this code inside csi-operator framework makes this PR non-trivial (you can see many commits in this PR on top of the first commit). The benefit of merging this PR with pre-defined assets in `generated` dir is to prove that the code really works and nothing overlooked.

The next PR auto-generating assets will have a benefit -- we can check how it works compiler-time without any cloud involved, just by comparing pre-defined assets with results of `hack/update-generated-assets.sh`. That PR may be complex, but very easy to test.

pkg/driver/gcp-pd/gcp_pd.go
```diff
@@ -0,0 +1,320 @@
...
+		c.GetConfigMapInformer(c.GuestNamespace),
+	)
+	informers := []factory.Informer{
+		c.GetConfigMapInformer(c.GuestNamespace).Informer(),
+	}
+	return hook, informers
+}
+
+// withCustomResourceTags adds resource tags from infrastructure.status.platformStatus.gcp.resourceTags to the
+// driver command line as --resource-tags=<parent_id>/<tagKey_shortname>/<tagValue_shortname>,...
```
rhrmo: looks like a typo, instead of `--resource-tags` it should say `--extra-tags`

rhrmo:
other than the comment typo it looks good to me

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-csi

mpatlasov:
/payload-job periodic-ci-openshift-release-main-nightly-5.0-e2e-gcp-ovn-serial

mpatlasov:
/verified by mpatlasov
I installed cluster with ClusterBot command `launch 5.0,openshift/csi-operator#576 gcp` and verified that driver deployment and daemonset are up and running and that each asset yaml file has corresponding object on api-server functionally the same as the object on a cluster before this PR.

PR #578, opened 2026-07-15
OCPBUGS-97920: allow specific actions for AWS EFS CredentialsRequest

dobsonj:
https://redhat.atlassian.net/browse/OCPBUGS-97920

This PR replaces the wildcard `elasticfilesystem:*` in the CredentialsRequest for AWS EFS with specific actions from https://github.com/openshift/aws-efs-csi-driver/blob/master/docs/iam-policy-example.json

/cc @openshift/storage

### DIFF
--- a/assets/overlays/aws-efs/base/credentials.yaml
+++ b/assets/overlays/aws-efs/base/credentials.yaml
@@ -13,7 +13,13 @@ spec:
     statementEntries:
     - effect: Allow
       action:
-      - elasticfilesystem:*
+      # https://github.com/openshift/aws-efs-csi-driver/blob/master/docs/iam-policy-example.json
+      - elasticfilesystem:DescribeAccessPoints
+      - elasticfilesystem:DescribeFileSystems
+      - elasticfilesystem:DescribeMountTargets
+      - elasticfilesystem:CreateAccessPoint
+      - elasticfilesystem:DeleteAccessPoint
+      - elasticfilesystem:TagResource
       - ec2:DescribeAvailabilityZones # for one-zone volumes
       resource: "*"
   # STS fields will be set by the operator
--- a/assets/overlays/aws-efs/generated/standalone/credentials.yaml
+++ b/assets/overlays/aws-efs/generated/standalone/credentials.yaml
@@ -15,7 +15,12 @@ spec:
     kind: AWSProviderSpec
     statementEntries:
     - action:
-      - elasticfilesystem:*
+      - elasticfilesystem:DescribeAccessPoints
+      - elasticfilesystem:DescribeFileSystems
+      - elasticfilesystem:DescribeMountTargets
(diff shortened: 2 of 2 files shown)


gnufied:
/testwith openshift/csi-operator/release-4.23/aws-efs-operator-e2e openshift/aws-efs-csi-driver#132 openshift/aws-efs-utils#46

gnufied:
(Don't mind my last command, I am just testing if this command works from csi-opeator repo)

assets/overlays/aws-efs/base/credentials.yaml
```diff
@@ -13,7 +13,13 @@ spec:
...
-      - elasticfilesystem:*
+      # https://github.com/openshift/aws-efs-csi-driver/blob/master/docs/iam-policy-example.json
+      - elasticfilesystem:DescribeAccessPoints
+      - elasticfilesystem:DescribeFileSystems
+      - elasticfilesystem:DescribeMountTargets
+      - elasticfilesystem:CreateAccessPoint
+      - elasticfilesystem:DeleteAccessPoint
+      - elasticfilesystem:TagResource
       - ec2:DescribeAvailabilityZones # for one-zone volumes
       resource: "*"
```
rhrmo, line 24: I see that the policy example restricts the resources of create/delete access point and tag resource with 
```
"Condition": {
        "StringLike": {
          "aws:RequestTag/efs.csi.aws.com/cluster": "true"
        }
```
should we also use this for these actions?

rhrmo: Ah I see it was also coderabbits comment and it is resolved, I wonder, can users set their own resource tags and that is why it would now work? Or what is the reason? 🤔

RomanBednar:
@dobsonj This CredentialsRequest with asterisk is also used in docs for EFS STS setup - if we change it here this should be updated: https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/storage/using-container-storage-interface-csi#efs-sts_persistent-storage-csi-aws-efs

There's also the cross-account setup in the documentation which does not seem impacted or related - but maybe a good idea to double-check.

There's also a step in CI for STS setups that also has the asterisk, found this one so far: https://github.com/openshift/release/blob/4bba61f00d631ebf49ed84db582dca9c1ad6b133/ci-operator/step-registry/ipi/conf/aws/oidc-creds-provision/ipi-conf-aws-oidc-creds-provision-commands.sh#L119

There does not seem to be any test configured for aws oidc in csi-operator repo that would use this workflow (openshift-e2e-aws-manual-oidc-sts). That's quite surprising - there are others like Azure...maybe we should add it.

But there is a job for cross account setup that might be worth running here that *is* configured. Just to check for regressions: aws-efs-sts-cross-account-operator-e2e

PR #580, opened 2026-07-20
NO-ISSUE: Bump golang.org/x/crypto to v0.52.0

jkaurredhat:
## Summary

This PR updates `golang.org/x/crypto` from `v0.49.0` to `v0.52.0` to address **CVE-2026-46597** (Denial of Service via crafted AES-GCM packet decoder inputs in `golang.org/x/crypto/ssh`).

An incorrectly placed cast from bytes to int allowed for server-side panic in the AES-GCM packet decoder for well-crafted inputs.

**Note**: `golang.org/x/crypto/ssh` is not directly imported by the csi-operator (indirect dependency only), so the risk is LOW. Updating for security compliance.


jsafrane:
This has been fixed as part of https://github.com/openshift/csi-operator/pull/582/

PR #585, opened 2026-07-27
STOR-2997: Auto generate assets for gcp pd csi driver

mpatlasov:
https://redhat.atlassian.net/browse/STOR-2997

This PR consists of two commits:
 * 1st -- all manual changes (base and patch yamls, generator go files)
 * 2nd -- only automatic update of generated yaml files (no manual changes)


mpatlasov:
/hold

to fix CI failures

pkg/operator/prerequisites_test.go
```diff
@@ -126,13 +127,13 @@ func TestApplyPrerequisites(t *testing.T) {
 			}
 
 			if !tc.expectErr {
-				verifyAppliedAssets(t, ctx, kubeClient, tc.assetDir, tc.assetNames)
+				verifyAppliedAssets(t, ctx, kubeClient, tc.assetNames)
```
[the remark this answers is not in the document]
mpatlasov: @coderabbitai, applyPrerequisites compacts assetNames only by:
```
					assetNames = append(assetNames[:i], assetNames[i+1:]...)
```
in L52 of pkg/operator/prerequisites.go. append() creates new descriptor for the slice, so such changes are not visible for the caller of applyPrerequisites(), hence passing tc.assetNames directly won't rewrite the table input. Does it make sense?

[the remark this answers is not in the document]
mpatlasov: @coderabbitai, please give me solid example of "backing-array mutation" you mentioned in you last comment. Please use only append construction like `assetNames = append(assetNames[:i], assetNames[i+1:]...)` in your example. I did some testing and never notice any visible change on caller level. My experiments tell me that no "backing-array mutation" is possible if you only execute `assetNames = append(assetNames[:i], assetNames[i+1:]...)`. Please back you statement by example go file with a caller and a function called when such append results in "backing-array mutation" visible on caller side.

assets/overlays/gcp-pd/patches/privileged_role_get_nodes.yaml
jsafrane, line 1: I think this file is not used anywhere (but please double-check)

pkg/driver/gcp-pd/gcp_pd.go
```diff
@@ -58,6 +59,83 @@ const (
...
+				),
+				commongenerator.DefaultPodNetworkLivenessProbe.WithExtraArguments(
+					"--probe-timeout=3s",
+				),
+			},
+			Assets: commongenerator.DefaultControllerAssets.WithAssets(generator.StandaloneOnly,
+				"base/rbac/kube_rbac_proxy_role.yaml",
+				"base/rbac/kube_rbac_proxy_binding.yaml",
+				"base/rbac/prometheus_role.yaml",
+				"base/rbac/prometheus_binding.yaml",
```
jsafrane: both prometheus_role and prometheus_binding are already part of `DefaultNodeAssets`. Adding them to ControllerAssets will create them twice in a standalone cluster.

assets/overlays/gcp-pd/generated/standalone/node.yaml
```diff
@@ -1,193 +1,197 @@
...
+#
+
 apiVersion: apps/v1
+kind: DaemonSet
 metadata:
   name: gcp-pd-csi-driver-node
-  namespace: openshift-cluster-csi-drivers
-  annotations:
-    config.openshift.io/inject-proxy: csi-driver
-    config.openshift.io/inject-proxy-cabundle: csi-driver
```
jsafrane, line 8: These annotations disappeared and I think it's wrong. They're necessary to configure HTTP proxy + CA bundle in the driver container, in case a cluster uses custom ones.

assets/overlays/gcp-pd/generated/standalone/node.yaml
```diff
@@ -1,193 +1,197 @@
...
-            - --v=${LOG_LEVEL}
-            - --enable-storage-pools=true
-            - --node-name=$(KUBE_NODE_NAME)
-          env:
-            - name: CSI_ENDPOINT
-              value: unix:/csi/csi.sock
-            - name: KUBE_NODE_NAME
-              valueFrom:
-                fieldRef:
-                  fieldPath: spec.nodeName
```
jsafrane, line 53: This env. var disappeared. I grepped the driver code, it does not use KUBE_NODE_NAME, but the upstream sample deployment yaml files do. I'd prefer if we set it.

assets/overlays/gcp-pd/generated/standalone/node.yaml
jsafrane: And I think that's why CI fails. Node pods are crashlooping. But there is nothing interesting in logs, so who knows...

assets/overlays/gcp-pd/base/storageclass.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: standard-csi
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: pd.csi.storage.gke.io
```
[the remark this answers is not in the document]
mpatlasov: > Choose exactly one default StorageClass for the generated overlay.

Correct. That's already implemented in pkg/driver/gcp-pd/gcp_pd.go: getStorageClassFiles() returns either  regularFiles or gcpDedicatedFiles. There is no bug there.

assets/overlays/gcp-pd/patches/privileged_role_get_nodes.yaml
mpatlasov: This file is used in L133 of gcp_pd.go:
```
			AssetPatches: commongenerator.DefaultGuestAssetPatches.WithPatches(generator.StandaloneOnly,
				...
				"privileged_role.yaml", "overlays/gcp-pd/patches/privileged_role_get_nodes.yaml.patch",
			),
```
This is required because our generic `assets/base/rbac/privileged_role.yaml` doesn't have that `get nodes` rule.

pkg/driver/gcp-pd/gcp_pd.go
mpatlasov: Agreed, fixed. I guess we could improve `verify-generated-assets` target in the Makefile to check that each manifest.yaml doesn't have any intersections between sections `controllerStaticAssetNames` and `guestStaticAssetNames`. Such improvement would reveal similar issue in samba assets, and potentially in others drivers.

assets/overlays/gcp-pd/generated/standalone/node.yaml
mpatlasov: Agreed. That's pure overlook from my side. I compared `.spec.template.metadata.annotations` where I had to compare `.metadata.annotations`. Anyway, fixed.

assets/overlays/gcp-pd/generated/standalone/node.yaml
mpatlasov: We still set it, everything is fine. It's just github web-ui renders changes this way. If you scroll up, you can see in L44:
```
+        - name: KUBE_NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
```

assets/overlays/gcp-pd/patches/controller_add_driver.yaml
```diff
@@ -0,0 +1,71 @@
...
+              containerPort: 10301
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 10
+            timeoutSeconds: 3
+            periodSeconds: 10
+            failureThreshold: 5
```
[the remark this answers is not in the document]
mpatlasov: This would be out-of-scope of this PR. We try auto-generate node.yaml and controller.yaml to be functionally the same as in `legacy/gcp-pd-csi-driver-operator/assets`.

assets/overlays/gcp-pd/patches/controller_add_driver.yaml
```diff
@@ -0,0 +1,71 @@
...
+              mountPath: "/etc/cloud-sa"
+            - name: bound-sa-token
+              mountPath: /var/run/secrets/openshift/serviceaccount
+              readOnly: true
+          securityContext:
+            readOnlyRootFilesystem: true
+          resources:
+            requests:
+              memory: 50Mi
+              cpu: 10m
```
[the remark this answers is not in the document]
mpatlasov: This would be out-of-scope of this PR. We try auto-generate controller.yaml to be functionally the same as in `legacy/gcp-pd-csi-driver-operator/assets`.

pkg/operator/prerequisites_test.go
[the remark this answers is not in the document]
mpatlasov: OK, I agree. Fixed.

assets/overlays/gcp-pd/patches/privileged_role_get_nodes.yaml
jsafrane: that's `privileged_role_get_nodes.yaml.patch`, not `privileged_role_get_nodes.yaml`

jsafrane:
/test e2e-gcp
/test e2e-gcp-csi
something is still crashing

jsafrane:
found it, the ports in livenessprobe are somehow wrong.

>  kind: Event
>  count: 181
>  message: "Liveness probe error: Get \"http://10.0.128.4:10300/healthz\": dial tcp
>    10.0.128.4:10300: connect: connection refused\nbody: \n"

mpatlasov:
> found it, the ports in livenessprobe are somehow wrong.
> 
> > kind: Event
> > count: 181
> > message: "Liveness probe error: Get "[http://10.0.128.4:10300/healthz\](http://10.0.128.4:10300/healthz%5C)": dial tcp
> > 10.0.128.4:10300: connect: connection refused\nbody: \n"

Fixed now. The magic line is `host: 127.0.0.1`. Borrowed from https://github.com/openshift/csi-operator/pull/364 .

mpatlasov:
/verified by CI
Also verified by mpatlasov that:
 * Installing a cluster with ClusterBot command launch 5.0,openshift/csi-operator#585, driver deployment and daemonset are up and running
 * Comparing each asset before this PR vs. after, they are functionally the same (except port numbers in service.yaml which was discussed and agreed with team in Slack)

PR #595, opened 2026-08-07
OCPBUGS-101779: azure-disk: Disallow empty cloud config

mpatlasov:
https://redhat.atlassian.net/browse/OCPBUGS-101779

We always pass cloud config to controller and node pods via `AZURE_CREDENTIAL_FILE`, hence empty cloud config must never happen.

This PR doesn't fully resolve OCPBUGS-101779, but if something goes wrong, it is better to return error explicitly than crashes with nil pointer dereference.


mpatlasov:
Let's keep options in-sync with azure-file driver.

PR #603, opened 2026-08-19
OCPBUGS-112272: node ServiceMonitor serverName uses guest namespace on HyperShift

bryan-cox:
## What

On HyperShift, the node metrics ServiceMonitor that csi-operator creates in the **guest** cluster set its TLS `serverName` from the **control-plane** namespace (`${NAMESPACE}`) instead of the guest namespace. The name never matched the service serving certificate, so every node metrics scrape failed certificate verification, the node CSI driver metrics target was permanently down, and `TargetDown` fired continuously.

Impact is limited to observability — CSI provisioning and attach/detach are unaffected — but node CSI metrics are never collected and the permanently-failing target produces continuous alert noise.

## Root cause

The shared patch `assets/common/metrics/service_monitor_add_port.yaml.patch` hardcoded `${NAMESPACE}` in `serverName`. That patch is applied by `generateDriverMetricsService()` for **both** the controller monitor (where `${NAMESPACE}` is correct, since the controller runs control-plane side) and the node monitor (where it is wrong, since the node DaemonSet and its metrics Service run guest side).

In `pkg/driver/common/operator/replacer.go` the two namespace variables resolve to different values under HyperShift:
- `${NAMESPACE}` → control-plane namespace
- `${NODE_NAMESPACE}` → guest namespace

On standalone both resolve to `openshift-cluster-csi-drivers`, which is why the defect was invisible there.

## Fix

- Introduce a `${MONITOR_NAMESPACE}` template variable in the shared ServiceMonitor patch.
- Resolve it per service prefix in the generator (`monitorNamespaceVariable`): `${NODE_NAMESPACE}` for the node monitor, `${NAMESPACE}` for the controller monitor. Wired into both `generateDriverMetricsService` and `generateSidecarMetricsServices`.
- Regenerated assets via `make update`.

## Scope / safety

- **HyperShift-only behavior change.** On standalone, `${NAMESPACE}` and `${NODE_NAMESPACE}` both resolve to `openshift-cluster-csi-drivers`, so the rendered `serverName` is byte-for-byte identical to before — only the template variable changed.
- **Controller ServiceMonitors unchanged** across all drivers (still `${NAMESPACE}`).
- Fix is in shared generator code, so any driver enabling node metrics inherits it. azure-disk is currently the only driver that generates a node metrics ServiceMonitor.

## Testing

- Added regression tests at the generator seam (`pkg/generator/asset_generator_test.go`) asserting the node monitor resolves to `${NODE_NAMESPACE}` and the controller monitor to `${NAMESPACE}`, using the real embedded assets.
- `make update` is idempotent (no unintended diffs), `go build ./...` clean, `go test ./pkg/...` all pass.

## Acceptance criteria

- [x] Node ServiceMonitor `serverName` resolves to the guest namespace, matching the serving cert SANs.
- [x] Generated HyperShift node asset renders `serverName` from `${NODE_NAMESPACE}`.
- [x] Controller metrics ServiceMonitor unchanged (`${NAMESPACE}`).
- [x] Standalone generated assets remain functionally correct.
- [x] Assets regenerated via the generator/make target; fix covers all affected driver overlays via shared code.

### DIFF
--- a/assets/common/metrics/service_monitor_add_port.yaml.patch
+++ b/assets/common/metrics/service_monitor_add_port.yaml.patch
@@ -10,4 +10,4 @@
     scheme: https
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
-      serverName: ${ASSET_PREFIX}-${SERVICE_PREFIX}-metrics.${NAMESPACE}.svc
+      serverName: ${ASSET_PREFIX}-${SERVICE_PREFIX}-metrics.${MONITOR_NAMESPACE}.svc
--- a/assets/overlays/azure-disk/generated/hypershift/node_servicemonitor.yaml
+++ b/assets/overlays/azure-disk/generated/hypershift/node_servicemonitor.yaml
@@ -19,7 +19,7 @@ spec:
     scheme: https
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
-      serverName: azure-disk-csi-driver-node-metrics.${NAMESPACE}.svc
+      serverName: azure-disk-csi-driver-node-metrics.${NODE_NAMESPACE}.svc
   jobLabel: component
   selector:
     matchLabels:
--- a/assets/overlays/azure-disk/generated/standalone/node_servicemonitor.yaml
+++ b/assets/overlays/azure-disk/generated/standalone/node_servicemonitor.yaml
@@ -19,7 +19,7 @@ spec:
     scheme: https
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
-      serverName: azure-disk-csi-driver-node-metrics.${NAMESPACE}.svc
+      serverName: azure-disk-csi-driver-node-metrics.${NODE_NAMESPACE}.svc
   jobLabel: component
   selector:
     matchLabels:
--- a/pkg/generator/asset_generator.go
+++ b/pkg/generator/asset_generator.go
@@ -170,6 +170,20 @@ func (gen *AssetGenerator) generateDeployment() error {
 	return nil
 }
 
+// monitorNamespaceVariable returns the runtime namespace variable that the
+// ServiceMonitor's TLS serverName must resolve to for the given servicePrefix.
+//
+// The node metrics Service runs guest-side, so its serving certificate is issued
+// for the guest namespace (${NODE_NAMESPACE}). The controller and its metrics
+// Service run control-plane side, so they use ${NAMESPACE}. On HyperShift these
+// two namespaces differ; on standalone they are identical. See OCPBUGS-112272.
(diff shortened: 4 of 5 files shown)


jsafrane:
@bryan-cox have you tested this change? Is its result visible in CI? Please add `/verified by ` by yourself.

bryan-cox:
> @bryan-cox have you tested this change? Is its result visible in CI? Please add `/verified by ` by yourself.

Hey @jsafrane - @clebs was planning on verifying it but I don't know where he's at on this atm. I'll check in with him today.

clebs:
/verified by @clebs 

Manual test plan execution to verify the fix:

# Results for test plan 603

## Test Scenarios

### Scenario 1: Node ServiceMonitor serverName resolves to guest namespace
✅ Confirm the core fix — node metrics ServiceMonitor TLS serverName matches the guest namespace, not the control-plane namespace.

On the azure hosted cluster check the node metrics `serverName`:
```bash
~ oc get servicemonitor -n openshift-cluster-csi-drivers -o jsonpath='{range .items[*]}{.metadata.name}{": "}{.spec.endpoints[*].tlsConfig.serverName}{"\n"}{end}'

azure-disk-csi-driver-node-monitor: azure-disk-csi-driver-node-metrics.openshift-cluster-csi-drivers.svc
```


### Scenario 2: Controller ServiceMonitor serverName resolves to control-plane namespace
✅ Verify the controller ServiceMonitor was not regressed — it should still use the control-plane namespace.

On the AWS management cluster check the control-plane-metrics servicemonitor `serverName`:
```bash
~ oc get servicemonitor -n clusters-bclement-hc-az -o jsonpath='{range .items[*]}{.metadata.name}{": "}{.spec.endpoints[*].tlsConfig.serverName}{"\n"}{end}' | grep control-plane-metrics

monitor-ovn-control-plane-metrics: ovn-kubernetes-control-plane.clusters-bclement-hc-az.svc
```

### Scenario 3: Node metrics are actually being scraped
✅ Confirm the fix resolves the actual symptom — node CSI driver metrics were never collected because TLS verification failed.


On the azure hosted cluster check Prometheus targets for the node metrics endpoint:
```bash
~ oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- \
curl -s http://localhost:9090/api/v1/targets | \
jq -r '.data.activeTargets[] | select(.labels.job | contains("node-metrics")) | "\(.labels.job) \(.health)"'

azure-disk-csi-driver-node-metrics up
azure-disk-csi-driver-node-metrics up
```

### Scenario 4: No TargetDown alert firing for node CSI metrics
✅ Confirm the alert noise described in the bug is resolved.


On the azure hosted cluster check that no `TargetDown` alerts are firing:
```bash
oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- \
curl -s http://localhost:9090/api/v1/alerts | \
jq -r '.data.alerts[] | select(.labels.alertname | contains("TargetDown")) | select(.labels.job | ascii_downcase |
 contains("csi")) | "\(.labels.alertname) \(.state)"'

<empty output>
```

PR #605, opened 2026-08-24
OCPBUGS-113538: Enable SELinuxMount tests

jsafrane:
Enable SELinuxMount tests in all CSI drivers that are based on block volumes.

NFS and CIFS- based CSI drivers don't support SELinux.


jsafrane:
/verified by CI
I can see SELinuxMount tests running in e2e-azure-csi e2e-gcp-csi e2e-aws-csi
/retest-required
/jira refresh

PR #615, opened 2026-09-01
manila: set os-certAuthorityPath in csi-manila-secrets for CA discovery

mandre:
The upstream Manila CSI driver (cloud-provider-openstack) has its own OpenStack client that reads CA certificates from the os-certAuthorityPath field in the CSI secret parameters, or the cacert field in clouds.yaml. It does NOT read the OS_CACERT environment variable. Neither field was set, so the driver fell back to the system cert pool. In HyperShift, the system cert pool on the management cluster may not include the guest cluster's OpenStack CA, causing TLS verification failures:

    failed to create Manila v2 client: failed to authenticate: tls: failed to verify certificate: x509: certificate signed by unknown authority

Add os-certAuthorityPath to the csi-manila-secrets template with a ${MANILA_CA_CERT_PATH} placeholder. At operator startup, the ExtraReplacementsFunc reads the cloud config ConfigMap (cloud-provider-config in standalone, openstack-cloud-config in HyperShift) and replaces the placeholder with the CA cert mount path when the ConfigMap contains ca-bundle.pem, or with an empty string when no custom CA is present. An empty string causes the driver to skip the CA file and use the system trust store, which is the correct behavior for public CAs.

### DIFF
--- a/assets/overlays/openstack-manila/base/config_secret.yaml
+++ b/assets/overlays/openstack-manila/base/config_secret.yaml
@@ -7,3 +7,4 @@ stringData:
   os-cloud: openstack
   os-cloudsFile: /etc/openstack/clouds.yaml
   os-useClouds: "true"
+  os-certAuthorityPath: "${MANILA_CA_CERT_PATH}"
--- a/assets/overlays/openstack-manila/generated/hypershift/config_secret.yaml
+++ b/assets/overlays/openstack-manila/generated/hypershift/config_secret.yaml
@@ -10,6 +10,7 @@ metadata:
   name: csi-manila-secrets
   namespace: openshift-manila-csi-driver
 stringData:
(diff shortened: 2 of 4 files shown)


jparrill:
What do we need to move this forward, can I help?

mandre:
> What do we need to move this forward, can I help?

Apologies, I should have added a hold on this PR. We can't merge it right away, because we haven't tested it doesn't break the non custom CA path (we have no cloud where to test manila that doesn't have a custom CA cert). We also wanted to go with a potentially cleaner solution @stephenfin was working on, with:
- https://github.com/openshift/hypershift/pull/9467
- https://github.com/openshift/csi-operator/pull/617

/hold

mandre:
At least this PR proves that we get a GREEN CI once we're able to consume the custom CA.

mandre:
Closing in favor of https://github.com/openshift/csi-operator/pull/617.
/close

PR #373, opened 2025-03-26
OSASINFRA-3675: Rework authentication in Manila CSI Driver Operator

dkokkino:
- Updated manila assets to mount secret and configmap to node/controller
- Modified the secret generator to use different parameters in preparation for change to manila driver

PR #617, opened 2026-09-01
OCPBUGS-115269: openstack-manila: Consume CA cert from CCO secret

stephenfin:
Update both the operator itself and the assets for the CSI driver to reflect changes made to cluster storage operator (CSO) in https://github.com/openshift/cluster-storage-operator/pull/557 and cloud credential operator (CCO) in https://github.com/openshift/cloud-credential-operator/pull/850.

This was originally targeted in https://github.com/openshift/csi-operator/pull/359 but https://github.com/openshift/csi-operator/pull/373 has broken hypershift support, so this is now targeted as a bug.

Depends on https://github.com/openshift/hypershift/pull/9467


mandre (commented):
/lgtm

PR #618, opened 2026-09-03
OCPBUGS-120667: fix(gcp-pd): don't apply guest-cluster resources against the management cluster on HyperShift

ckandag:
Bug: OCPBUGS-120667

## Summary

Problem: On HyperShift, the operator applies `PrerequisiteAssets` using the management-cluster client. Several of those assets are guest-cluster resources that shouldn't be applied to the management cluster at all.

Fix: `GetGCPPDOperatorConfig` now takes an `isHypershift` flag. On HyperShift, `PrerequisiteAssets` only includes `controller_sa.yaml`. Among the rest:

- `hostnetwork_role.yaml` / `controller_hostnetwork_binding.yaml` — not needed at all on HyperShift, since the controller pod runs with `hostNetwork: false` there.
- `node_sa.yaml` / `privileged_role.yaml` / `node_privileged_binding.yaml` — still needed, these guest-cluster resources are already applied correctly via `GuestConfig.Assets`, so listing them here too was just duplicating.

Standalone is unaffected (`isHypershift=false` keeps the full list).

## Test plan

- [x] `go build ./...`
- [x] Live-verified on a HyperShift GCP HostedCluster: with this fix, `gcp-pd-csi-driver-operator` no longer crash-loops on `Forbidden` errors applying cluster-scoped RBAC against the management cluster.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cmd/gcp-pd-csi-driver-operator/main.go
+++ b/cmd/gcp-pd-csi-driver-operator/main.go
@@ -50,6 +50,6 @@ func NewOperatorCommand() *cobra.Command {
 }
 
 func runCSIDriverOperator(ctx context.Context, controllerConfig *controllercmd.ControllerContext) error {
-	opConfig := gcp_pd.GetGCPPDOperatorConfig()
+	opConfig := gcp_pd.GetGCPPDOperatorConfig(*guestKubeconfig != "")
 	return operator.RunOperator(ctx, controllerConfig, *guestKubeconfig, opConfig)
 }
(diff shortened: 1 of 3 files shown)


ckandag:
@rhrmo : Could you please help with this update as well?

I had missed including these changes in my initial PR. While merging the HyperShift portion of the work, the HyperShift E2E CI tests caught this bug.

I need this fix to go in for the HyperShift CI to succeed and for the HyperShift work to merge. I have re-verified the PV/PVC tests with all the changes across the relevant repositories (CSI, CSO, and HyperShift operators).

Hopefully, this is the last one in this PR.

rhrmo:
/cancel Approve
I will Approve again after a bug is filed and added to the PR

ckandag:
@rhrmo : i created a new bug , following the same linked pattern as the one you have createad (

earlier :   OCPBUGS-112748→5.1.0 / OCPBUGS-114002→5.0.z
this:  OCPBUGS-120667->5.1.0 / OCPBUGS-120666

Please let me know if that looks right

ckandag:
@rhrmo Can I pls get an approve if the oc-bug reference now looks good. All the failing tests looks like are optional ones

PR #625, opened 2026-09-17
[master] OCPBUGS-126031,OCPBUGS-126030: Bump google.golang.org/grpc to v1.82.2 to address CVE-2026-88031

sunku5494:


sunku5494:
The CVE fix is not required to bring in here as affected pkg grpc/xds is neither vendored nor imported

